PR #15, opened 2020-01-16
Prove dht

aslesarenko:


lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -29,7 +34,7 @@
         for (int i = 0;
              i < unspentBoxes.size() &&
                      (collected < amountToSpend ||
-                             (!tokenOpt.isPresent() || collectedTokens < tokenAmount)
+                             (tokenOpt.isPresent() && collectedTokens < tokenAmount)
```
greenhat: Hey, that's much better! Thanks!

PR #18, opened 2020-01-20
Update sigma (remove dex deadline)

greenhat:


common/src/main/java/org/ergoplatform/appkit/AppkitProvingInterpreter.scala
```diff
@@ -14,12 +14,12 @@ import org.ergoplatform.wallet.protocol.context.{ErgoLikeParameters, ErgoLikeSta
 import scala.util.Try
 import sigmastate.eval.CompiletimeIRContext
 import sigmastate.interpreter.{ContextExtension, ProverInterpreter}
-import sigmastate.utils.Helpers._  // don't remove, required for Scala 2.11
 
 import scala.collection.mutable
 
 object Helpers {
   implicit class AppkitTryOps[A](val source: Try[A]) extends AnyVal {
+    import sigmastate.utils.Helpers._  // don't remove, required for Scala 2.11
```
greenhat: @aslesarenko I moved import inside the class due to "ambiguous implicits" error on 2.12.

aslesarenko: this class was temporary until sigma is updated. This was the reason for ambiguous implicits. :)

PR #21, opened 2020-01-25
Update ProveDHTuple to use 4 parameters

scalahub:


lib-api/src/main/java/org/ergoplatform/appkit/ErgoProverBuilder.java
```diff
@@ -31,28 +33,22 @@
...
+     * @return
      *
-     * @param otherParty address of the other party whose public key is to be included in
-     *                   Diffie-Hellman Tuple. This prover is assumed to belong to Bob, then `otherParty`
-     *                   is Alice.
-     * @see
-     * <a href="https://github.com/ScorexFoundation/sigmastate-interpreter/blob/b3695bdb785c9b3a94545ffea506358ee3f8ed3d/sigmastate/src/test/scala/sigmastate/utxo/examples/DHTupleExampleSpecification.scala#L28">example</a>
      */
-    ErgoProverBuilder withSecondDHTSecret(Address otherParty);
+    ErgoProverBuilder withDHTData(GroupElement g, GroupElement h, GroupElement u, GroupElement v, BigInteger x);
```
aslesarenko: Isn't it that the generator is always the same?

lib-api/src/main/java/org/ergoplatform/appkit/ErgoProverBuilder.java
```diff
@@ -31,28 +33,22 @@
     ErgoProverBuilder withSecretStorage(SecretStorage storage);
 
     /**
-     * Add DHT prover input using this prover's secret as belonging to the first party (i.e. Alice).
+     * ProveDHTuple is of the form (g, h, u, v) with secret x (and unknown y), where:
+     *   h = g^y
+     *   u = g^x
+     *   v = g^xy
```
aslesarenko: This is quite cryptic for people like me. 
In order to understand it I had to went through the original example etc.
So I suggest to leave the original naming and add this crypto related information (not replace the original docs)

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ErgoProverBuilderImpl.java
```diff
@@ -23,8 +21,7 @@
     private final BlockchainContextImpl _ctx;
 
     private ExtendedSecretKey _masterKey;
-    private DiffieHellmanTupleProverInput _firstSecret;
-    private DiffieHellmanTupleProverInput _secondSecret;
+    private DiffieHellmanTupleProverInput _dhtSecret;
```
aslesarenko: I assumed the proved can have both secrets, don't think we should limit the API to allow provers with single secret.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ErgoProverBuilderImpl.java
```diff
@@ -49,34 +46,16 @@ public ErgoProverBuilder withSecretStorage(SecretStorage storage) {
...
     /**
-     * This code uses original cryptographic notation known from the literature, see the following
-     * <a href="https://github.com/ScorexFoundation/sigmastate-interpreter/blob/b3695bdb785c9b3a94545ffea506358ee3f8ed3d/sigmastate/src/test/scala/sigmastate/utxo/examples/DHTupleExampleSpecification.scala#L28">example</a>
-     * to understand the implementation, where this _masterKey belong to Bob
+     * ProveDHTuple is of the form (g, h, u, v) with secret x (and unknown y), where:
+     *   h = g^y
+     *   u = g^x
+     *   v = g^xy
+     *
+     *   NOTE: We can swap x, y and obtain another tuple (g, u, h, v) with secret y (and unknown x).
```
aslesarenko: Again you can add this crypto terminology in addition to the original names and docs.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/InputBoxImpl.java
```diff
@@ -63,6 +64,11 @@ public String toJson(boolean prettyPrint) {
         return json;
     }
 
+    @Override
+    public Map<ErgoBox.NonMandatoryRegisterId, ? extends Values.EvaluatedValue<? extends SType>> getRegisters() {
```
aslesarenko: This sigma types should not be exposed to Java interfaces. And returning registers as Map doesn't really make sense, since they are actually packed into dense array. So I suggest this method to return `List<ErgoValue<?>>`.
To also be consistent with `OutBoxBuilder registers(ErgoValue<?>... registers)`.

lib-api/src/main/java/org/ergoplatform/appkit/ErgoProverBuilder.java
scalahub: In feel it should not be the same generator all the time. One variant of ErgoMix uses proveDHTuple(h, h, v, v), where h=g^y and v = g^(xy). Basically this allows us do to proveDlog using a generator different from g. That is: proveDlog_h(v) = proveDHTuple(h, h, v, v).

Since the protocol allows changing the generator, l think we should also allow it.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ErgoProverBuilderImpl.java
scalahub: I think (but hopefully others can chime in), that proveDHtuple makes sense for proving knowledge of only one of x or y, since the protocol will use only one of the secret even if the prover knows both. For proving knowledge of both, the prover must do two proveDlogs, one for each secret. Also, as far as I understand, the underlying sigma does not support two secrets for proveDHtuple.

lib-api/src/main/java/org/ergoplatform/appkit/ErgoProverBuilder.java
scalahub: I agree. We should put the original info too.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ErgoProverBuilderImpl.java
scalahub: Agree (I removed the original test by mistake). We need to put it back.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/InputBoxImpl.java
scalahub: I believe this is addressed in PR #23

scalahub:
By the way, a more suitable example for proveDHTuple is 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/291997a1a2330ebe0ae5ea5a7c15810e6a8deae7/sigmastate/src/test/scala/sigmastate/utxo/examples/MixExampleSpecification.scala#L56
and
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/291997a1a2330ebe0ae5ea5a7c15810e6a8deae7/sigmastate/src/test/scala/sigmastate/utxo/examples/MixExampleSpecification.scala#L86

lib-impl/src/main/java/org/ergoplatform/appkit/impl/InputBoxImpl.java
aslesarenko: That is strange that I don't see my getRegisters method here.

common/src/main/java/org/ergoplatform/appkit/JavaHelpers.scala
```diff
@@ -23,10 +23,10 @@ import java.lang.{Long => JLong, String => JString}
 import java.math.BigInteger
 import java.util.{List => JList}
 
-import sigmastate.utils.Helpers._  // don't remove, required for Scala 2.11
+import sigmastate.utils.Helpers._
```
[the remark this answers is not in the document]
scalahub: fixed

PR #22, opened 2020-01-26
Expose requests to get spending transaction for a box (by id)

greenhat:
Expose API requests needed to implement https://github.com/aslesarenko/ergo-tool/issues/31

Todo: 
- [x] get `transactions/boxes/id` and return `InputBox` (with added `spendingTxId`);
- [ ] get `transactions/id` and return [`FullTransaction`](http://github.com/aslesarenko/ergo-appkit/blob/00ce6c27b27631d6bf5aade621a05791096ab3bc/java-client-generated/src/main/java/org/ergoplatform/explorer/client/model/FullTransaction.java#L35-L35) as  ???;

### DIFF
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
@@ -3,6 +3,7 @@ package org.ergoplatform.appkit;
 import sigmastate.Values;
 
 import java.util.List;
+import java.util.Optional;
 
 /**
  * This interface represent a specific context of blockchain for execution
@@ -67,5 +68,8 @@ public interface BlockchainContext {
 
     /** Get unspent boxes protected by given ergo tree template */
     List<InputBox> getUnspentBoxesForErgoTreeTemplate(ErgoTreeTemplate template);
+
+    Optional<InputBox> getBoxByIdFromExplorer(ErgoId boxId);
+
 }
 
--- a/lib-api/src/main/java/org/ergoplatform/appkit/InputBox.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/InputBox.java
@@ -3,6 +3,7 @@ package org.ergoplatform.appkit;
 import sigmastate.Values;
 
 import java.util.List;
+import java.util.Optional;
 
 /**
  * Interface of UTXO boxes which can be accessed in the blockchain node.
@@ -32,5 +33,7 @@ public interface InputBox {
     Values.ErgoTree getErgoTree();
 
     String toJson(boolean prettyPrint);
+
+    Optional<ErgoId> getSpentTransactionId();
 }
 
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
@@ -9,8 +9,7 @@ import retrofit2.Retrofit;
 import scorex.util.encode.Base16;
 import sigmastate.Values;
 
-import java.util.ArrayList;
-import java.util.List;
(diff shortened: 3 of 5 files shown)


lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
```diff
@@ -67,5 +68,8 @@
 
     /** Get unspent boxes protected by given ergo tree template */
     List<InputBox> getUnspentBoxesForErgoTreeTemplate(ErgoTreeTemplate template);
+
+    Optional<InputBox> getBoxByIdFromExplorer(ErgoId boxId);
+
```
aslesarenko: Why `InputBox[] getBoxesById(String... boxIds) throws ErgoClientException;` is not enough?

lib-api/src/main/java/org/ergoplatform/appkit/InputBox.java
```diff
@@ -32,5 +33,7 @@
     Values.ErgoTree getErgoTree();
 
     String toJson(boolean prettyPrint);
+
+    Optional<ErgoId> getSpentTransactionId();
```
aslesarenko: What this method should do? What does it mean "spent transaction"? Transaction cannot be spent.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
```diff
@@ -149,5 +149,17 @@ public ErgoContract compileContract(Constants constants, String ergoScript) {
...
+    @Override
+    public Optional<InputBox> getBoxByIdFromExplorer(ErgoId boxId) {
+        Optional<TransactionOutput> res;
+        try {
+            res = Optional.ofNullable(ExplorerFacade
+                    .transactionsBoxesIdGet(_retrofitExplorer, boxId.toString()));
+        } finally {
+            res = Optional.empty();
+        }
+        return res.flatMap(out -> getInputBoxes(Collections.singletonList(out)).stream().findFirst());
```
aslesarenko: This doesn't seem right, all this logic should happen on the explorer side, so that unnecessary data is not leaving postgres.

lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
greenhat: `InputBox[] getBoxesById(String... boxIds) throws ErgoClientException;` uses node's `utxo/byId/{boxId}` returning `ErgoTransactionOutput` and I need to call explorer's `transactions/boxes/{id}` returning `TranscationOutput` because I need it's `spentTransactionId`id of the tx where this box was used as input.

lib-api/src/main/java/org/ergoplatform/appkit/InputBox.java
greenhat: It's an id of the transaction where the box was used as input. See http://github.com/ergoplatform/explorer-back/blob/73096d2e630609bcee0de1b4cb42ad8d81e445d1/src/main/resources/api/openapi.yaml#L866-L866
But seems like it does not belong to `InputBox` at all. We need a type to represent `TransactionOutput` with a more generic class. Maybe `TxOutBox`?

lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
greenhat: Well, I don't like exposing this inner machinery either. Maybe it should be thrown away in favor of https://github.com/ergoplatform/explorer-back/issues/232.

greenhat:
Closed in favor of new backend API call - https://github.com/ergoplatform/explorer-backend/issues/3

PR #31, opened 2020-02-10
fallback to sigma 3.1.1

greenhat:


greenhat:
@aslesarenko Besides the "old" ergo wallet version `appkit-wallet-f7f7d673-SNAPSHOT` it's ready to merge. I could not find the branch from which this ergo wallet version was published to check what sigma version is used there.

PR #32, opened 2020-02-11
Design contracts

aslesarenko:


docs/design-contracts/design-contracts.md
```diff
@@ -0,0 +1,412 @@
...
+
+Let's take the output created by the arrow labeled with `buyerOut@0`. This label tell us
+that the output is at the index `0` in the `OUTPUTS` collection of the transaction and that
+we can refer to this box by the `buyerOut` name. Thus we can label both the box itself and
+the arrow to give it a name.
+
+The conditions shown in the `buyerOut` box have the form `bid ? condition`, which means
+they should be checked before the `bid` box can be spent. 
+The conditions have the following meaning:
+1) `tAmt: TID` require the box to have `tAmt` amount of `TID` token
```
greenhat: @aslesarenko Initially I confuse `tAmt` from `bid ? tAmt: TID` with `tAmt` from `SellOrder`. I suggest exposing some contract internals (like token amount) inside the contract box on the diagram and give them different names.

greenhat:
It'd be nice to see diagram diffs and previews. Here is a list of possible alternatives for PP - https://gist.github.com/rodneyrehm/40e7946c0cff68a31cea

docs/design-contracts/design-contracts.md
```diff
@@ -331,6 +331,78 @@ can be created when there are two matching boxes with `buyOrder` and `sellOrder`
...
+
+After transformation we can get the correct script which checks all the required
+preconditions for spending the `buyOrder` box.
+```
+/** buyOrder contract */
+{
+  val cancelCondition = { buyer }
+  val swapCondition = try {
+    val buyerOut = OUTPUTS(0)                     // from buyerOut@0
+    val tid = buyerOut.tokens(TID)
```
greenhat: @aslesarenko I thought `TID` is a token id(byte array), here it's used as an index.

docs/design-contracts/design-contracts.md
```diff
@@ -331,6 +331,78 @@ can be created when there are two matching boxes with `buyOrder` and `sellOrder`
...
+3) `@contract`  ==> `buyerOut.propositionBytes == buyer.propBytes` where `R4 = buyerOut.R4[Coll[Byte]].get` 
+4) `buyerOut@0` ==> `val buyerOut = OUTPUTS(0)`
+
+After transformation we can get the correct script which checks all the required
+preconditions for spending the `buyOrder` box.
+```
+/** buyOrder contract */
+{
+  val cancelCondition = { buyer }
+  val swapCondition = try {
```
greenhat: @aslesarenko I think it would hard to convince verifier to close their eyes on certain condition failures inside `try` function. I'd rather introduce explicit `require(conditions)` since it translates into verifier's pre-conditions automatically and we can just ignore on evaluation. I see forcing a user to write requirements beforehand as an upside here because a user must be ready to all possible scenarios.

docs/design-contracts/design-contracts.md
aslesarenko: I should probably make is more clear. tAmt is not a field, it is a parameter of the diagram - Long value of tokens. Or put it another way, it is a shared parameter both for Sell Order and Swap transactions. So basically same name means the same value throughout the diagram. However, only explicit conditions marked with `?` are transformed to ErgoScript. But all others are actually ensured during transaction building (Appkit code)

docs/design-contracts/design-contracts.md
aslesarenko: Yes, I rejected the `try` idea and instead changes the code using nested && operations.

docs/design-contracts/design-contracts.md
aslesarenko: Yes, in the diagram it is id, but we don't have `tokens.getByKey(TID)` operation, so then the diagram is translated into ErgoScript, TID is provided as named constant. The correspondence between tokenId, TID constant and real content of the buyerOut box is ensured buy the tx builder code in Appkit, this is possible since all this parameters are created there.

greenhat:
@aslesarenko I was thinking of some pseudo-language to describe what is happening in "DEX matching (simplified)" diagram without drawing stuff. Here is a super raw take on how it might look like:
```
def dex(buyer: PK, seller: PK, E: ERG, ergAmt: ERG, tAmt: TID) = {

  val buyerWallet = Wallet(template = pk(buyer))
  val BuyerContract = ...
  val buyerBalanceBoxes = buyerWallet.selectBoxes(toSpent = ergAmt)
  val buyerBidBox = Box(value = ergAmt, script = BuyerContract)
  val buyerChangeBox = Box(value = buyerBalanceBoxes.totalValue - ergAmt, script = pk(buyer))
  val buyOrderTransaction = Transaction(
    inputs = List(buyerBalanceBoxes),
    outputs = List(buyerBidBox, buyerBidChange)
  )

  val sellerWallet = Wallet(template = pk(seller))
  val SellerContract = ...
  val sellerBalanceBoxes = sellerWallet.selectBoxes(ergToSpent = minErg, tokenToSpend = (tAmt: TID))
  val sellerAskBox = Box(value = minErg, tokens = (tAmt: TID), script = SellerContract)
  val sellerChangeBox = Box(
    value  = sellerBalanceBoxes.totalValue - minErg,
    tokens = (sellerBalanceBoxes.totalTokens - tAmt: TID),
    script = pk(seller)
  )
  val sellOrderTransaction = Transaction(
    inputs = List(sellerBalanceBoxes),
    outputs = List(sellerAskBox, sellerChangeBox)
  )

  val buyerRefundBox = Box(value = ergAmount, tokens = (1: MintedTID), script = pk(buyer))
  val cancelBuyTransaction = Transaction(inputs = List(buyerBidBox), outputs = List(buyerRefundBox))

  val sellerRefundBox = Box(value = minErg, tokens = (tAmt: TID), script = pk(seller))
  val cancelSellTransaction = Transaction(inputs = List(sellerAskBox), outputs = List(sellerRefundBox))

  val buyerOutBox = Box(value = minErg, tokens = (tAmt: TID), script = pk(buyer))
  val sellerOutBox = Box(value = ergAmt, script = pk(seller))
  val swapTransaction = Transaction(
    inputs = List(buyerBidBox, sellerAskBox),
    outputs = List(buyerOutBox, sellerOutBox)
  )
}
```
Areas of future development:
- generate a diagram (starting with simple arrows for through transaction's input and outputs);
- implement as DSL and/or macro in Scala and enable tighter integration with [devepment workflow](https://github.com/ScorexFoundation/sigmastate-interpreter/pull/610).

aslesarenko:
@greenhat Seems, you still think the diagram is just a picture. I will make it more clear that the diagram is just a view of the underlying spec :)

For your example.
1) It misses the point of specifying conditions like `bid ? R4 == bid.id` in the boxes, rather than in the contracts as we used to think. I don't see them at all in the example, however it is the key idea. 
If you try to add them, you will stumble upon the same old problem of Deep DSL Embedding in Scala: how to represent expressions? Macroses are too heavy weight.

2) Yes we can think of some Embedded DSLs, but I want to go one step further and specify FlowCards Specification EIP and a standardized file format (Json/XML/Protobuf) which will be language agnostic. 

3) Visual notation can be generated, atleast as a workaround until ideally we have a graphical editing tools. Having a diagram it is easy to write the text of `def dex(...) = ...` function, but it is much harder to write the code first, without having a diagram on the screen. So the first tooling we need is a diagram generator.

4) Whatever tools are used (DSL, Diagram Editor, etc) they all will produce a component which I call FlowCard (e.g. file `*.flowcard` specification format) (or whatever storage we use) which will be used for execution.

5) You can think about how how to generate input for stainless validation, but I think this will make sense after I finish with FlowCard semantic specification.

aslesarenko:
> eliminate multi page explanation which needed for a diagram.
Those explaination is just an example, once you understand semantics of FlowCard and diagrams, no need to explain anything, it can be either manually translated to the code (e.g. Appkit based) or just run directly by FlowCards runtime.

greenhat:
> @greenhat Seems, you still think the diagram is just a picture. I will make it more clear that the diagram is just a view of the underlying spec :)
> 
> For your example.
> 
> 1. It misses the point of specifying conditions like `bid ? R4 == bid.id` in the boxes, rather than in the contracts as we used to think. I don't see them at all in the example, however it is the key idea.
>    If you try to add them, you will stumble upon the same old problem of Deep DSL Embedding in Scala: how to represent expressions? Macroses are too heavy weight.

Good catch! I just missed them. Here are them(changed lines):  
```
  val buyerOutBox = Box(value = minErg, tokens = (tAmt: TID), registers = (R4 -> buyerBidBox.id), script = pk(buyer))
  val sellerOutBox = Box(value = ergAmt, registers = (R4 -> sellerAskBox.id), script = pk(seller))
``` 

I'm looking forward to FlowCard spec!

PR #34, opened 2020-02-13
add OutBoxBuilder.mintToken

greenhat:
for https://github.com/aslesarenko/ergo-tool/pull/35

### DIFF
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
@@ -7,6 +7,8 @@ import scala.Tuple2;
 import sigmastate.Values;
 import special.collection.Coll;
 
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 
@@ -46,10 +48,11 @@ public class OutBoxBuilderImpl implements OutBoxBuilder {
     }
 
     @Override
-    public OutBoxBuilder mintToken(ErgoToken token, String tokenName, String tokenDescription, String tokenNumberOfDecimals) {
-        ErgoValue<Coll<Byte>> tokenNameVal = ErgoValue.of(tokenName.getBytes());
-        ErgoValue<Coll<Byte>> tokenDescVal = ErgoValue.of(tokenDescription.getBytes());
-        ErgoValue<Coll<Byte>> tokenNumOfDecVal = ErgoValue.of(tokenNumberOfDecimals.getBytes());
+    public OutBoxBuilder mintToken(ErgoToken token, String tokenName, String tokenDescription, int tokenNumberOfDecimals) {
+        Charset utf8 = StandardCharsets.UTF_8;
+        ErgoValue<Coll<Byte>> tokenNameVal = ErgoValue.of(tokenName.getBytes(utf8));
+        ErgoValue<Coll<Byte>> tokenDescVal = ErgoValue.of(tokenDescription.getBytes(utf8));
+        ErgoValue<Coll<Byte>> tokenNumOfDecVal = ErgoValue.of(Integer.toString(tokenNumberOfDecimals).getBytes(utf8));
         Collections.addAll(_registers, tokenNameVal, tokenDescVal, tokenNumOfDecVal);
         Collections.addAll(_tokens, token);
         return this;
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
@@ -5,6 +5,7 @@ import org.ergoplatform.*;
 import org.ergoplatform.appkit.*;
 import scala.Tuple2;
 import sigmastate.Values;
+import special.collection.Coll;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
```diff
@@ -44,6 +45,16 @@ public OutBoxBuilderImpl tokens(ErgoToken... tokens) {
...
 
+    @Override
+    public OutBoxBuilder mintToken(ErgoToken token, String tokenName, String tokenDescription, String tokenNumberOfDecimals) {
+        ErgoValue<Coll<Byte>> tokenNameVal = ErgoValue.of(tokenName.getBytes());
+        ErgoValue<Coll<Byte>> tokenDescVal = ErgoValue.of(tokenDescription.getBytes());
+        ErgoValue<Coll<Byte>> tokenNumOfDecVal = ErgoValue.of(tokenNumberOfDecimals.getBytes());
+        Collections.addAll(_registers, tokenNameVal, tokenDescVal, tokenNumOfDecVal);
+        Collections.addAll(_tokens, token);
+        return this;
+    }
```
aslesarenko: @greenhat the Appkit API should be user centric. It means it should incapsulate as much complexity as it make sense. That is what builder is for. In particular:
1) Token id should be inferable automatically
2) The TX builder should check that only one token is minted
3) tokenNumberOfDecimals should be of type Int

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
```diff
@@ -44,6 +45,16 @@ public OutBoxBuilderImpl tokens(ErgoToken... tokens) {
         return this;
     }
 
+    @Override
+    public OutBoxBuilder mintToken(ErgoToken token, String tokenName, String tokenDescription, String tokenNumberOfDecimals) {
+        ErgoValue<Coll<Byte>> tokenNameVal = ErgoValue.of(tokenName.getBytes());
```
aslesarenko: I think encoding should be explicit UTF-8 or how it is specified in EIP4

greenhat: Good catch! Thanks! Fixed.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.java
greenhat: > 1. Token id should be inferable automatically

Since token id for minted token should be the first input id of the transaction, I don't see a nice way to infer it in`OutBoxBuilder.mintToken`. The only way I see is to require all tx inputs are specified before `OutBoxBuilder.mintToken` call which is confusing.

> 2. The TX builder should check that only one token is minted

Done.

> 3. tokenNumberOfDecimals should be of type Int

Done.

aslesarenko: This where the builder magic come into play. You can remember the parameters and then perform actual minting then the final TransactionBuilder.build() is called, at which point all the information should be present.
Think of the build as a transaction compiler which incapsulate all the internal complexity.

greenhat: I've considered a few options and the best I've got is a new method in `UnsignedTransactionBuilder`:
```
    UnsignedTransactionBuilder outputWithMintedToken(OutBox output,
                                                     long tokenAmount,
                                                     String tokenName,
                                                     String tokenDescription,
                                                     int tokenNumberOfDecimals);
```
But I still don't like it. It'd require adding token and registers to already "built" outbox, adding "rebuilding" method in `OutBoxBuilder`. I don't find this change that much helpfull to the user compared to the original solution(`OutBoxBuilder.mintToken`). 
Could you give me a hint for a better implementation?

aslesarenko: 1) You can add internal method `mintToken` to UnsignedTransactionBuilderImpl
2) From existing `public OutBoxBuilder mintToken(...` you can call `_txB.mintToken(...)` and pass parameters to tx builder
3) in UnsignedTransactionBuilderImpl.mintToken check it is not called twice
4) in UnsignedTransactionBuilderImpl.build access first input ErgoBox in _inputBoxes to get it's id
5) in UnsignedTransactionBuilderImpl.build use parameters to add necessary token to the output box with id from prev step.
6) validate information on every step.

greenhat: > 1. You can add internal method `mintToken` to UnsignedTransactionBuilderImpl
> 2. From existing `public OutBoxBuilder mintToken(...` you can call `_txB.mintToken(...)` and pass parameters to tx builder

How to identify the output box here? It does not exist yet, but tx builder have to add token to it later. Add a syntetic id of the building box in box builder?

> 5. in UnsignedTransactionBuilderImpl.build use parameters to add necessary token to the output box with id from prev step.

Here we need to know the output box.

aslesarenko: Ok, you can implement `mintToken` as a method of tx builder and call it when all information is already available, i.e. immediately before build().
You can pass output index to which the new token will be added as an additional argument of mintToken.

greenhat: That would introduce a requirement for certain tx build methods to be called in a specific order. Also, adding a token after a box is built does not look nice(tokens order assumptions, etc.).  
The more I think of it, the more I come to the conclusion that having `mintToken` in tx builder might do more harm than good.
Starting from the overall idea of "hiding" token id from the caller. I believe in some cases caller **wants** to know the minted token id.
Besides that, by explicitly requiring parameters to the token minting process(token id, etc.) we force the user to understand how token minting works.

aslesarenko: ok, let's expose this details for now.

PR #38, opened 2020-03-03
extract building of UnsignedErgoLikeTransaction to use in playgrounds

greenhat:


lib-impl/src/main/java/org/ergoplatform/appkit/impl/TransactionOperations.scala
```diff
@@ -0,0 +1,61 @@
...
+    )
+
+    val tx = new UnsignedErgoLikeTransaction(
+      inputs.map(b => new UnsignedInput(b.id)),
+      dataInputs,
+      finalOutputCandidates.toIndexedSeq
+    )
+    tx
+  }
+}
```
aslesarenko: @greenhat this code looks like belonging to sigmastate library, i.e. it mostly uses Sigma classes.
But I wonder how many of such methods we will need on Sigma side? There needs to be some systematic approach for creating them. And also how to make this methods stable, i.e. they should cover all possible transactions, withing our underlying transaction model.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.java
```diff
@@ -85,53 +86,16 @@ public UnsignedTransactionBuilder sendChangeTo(ErgoAddress changeAddress) {
...
-        UnsignedErgoLikeTransaction tx =
-                new UnsignedErgoLikeTransaction(inputs, dataInputs, outputCandidates);
-        List<ErgoBox> boxesToSpend =
-                _inputBoxes.stream().map(b -> b.getErgoBox()).collect(Collectors.toList());
+        IndexedSeq<ErgoBox> inputBoxes = JavaHelpers.toIndexedSeq(boxesToSpend);
+        UnsignedErgoLikeTransaction tx = TransactionOperations.buildUnsignedErgoTx(inputBoxes, dataInputs,
+                outputCandidates, _feeAmount, _changeAddress, _ctx.getHeight());
         ErgoLikeStateContext stateContext = createErgoLikeStateContext();
 
         return new UnsignedTransactionImpl(tx, boxesToSpend, new ArrayList<>(), stateContext);
```
aslesarenko: @greenhat As long as `TransactionOperations.buildUnsignedErgoTx` is moved to Sigma this code is ok.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/TransactionOperations.scala
greenhat: @aslesarenko I'd rather refrain from adding stuff to sigma, that sigma itself does not need. It's already huge.
We can put them into separate lib, or add to the wallet(we already have various box operations there).

aslesarenko: Ah, yes wallet is better. It looks also related to the issue https://github.com/ergoplatform/ergo/issues/1029

greenhat:
Created #44

greenhat:
Done in #46

PR #50, opened 2020-04-15
Add support for multi ProveDHTuple and ProveDLog

scalahub:


appkit/src/test/scala/org/ergoplatform/appkit/DHTProverSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+  with HttpClientTesting {
+
+  property("DHTProver") {
+    val ergoClient = createMockedErgoClient(MockData(Nil, Nil))
+    val g: GroupElement = CryptoConstants.dlogGroup.generator
+    val x = BigInt("187235612876647164378132684712638457631278").bigInteger
+    val y = BigInt("340956873409567839086738967389673896738906").bigInteger
+    val gX:GroupElement = g.exp(x)
+    val gY:GroupElement = g.exp(y)
+    val gXY:GroupElement = gY.exp(x)
```
aslesarenko, line 24: @scalahub shouldn't it be `gX.exp(y)`?

appkit/src/test/scala/org/ergoplatform/appkit/DHTProverSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+          |  proveDHTuple(gY, gY, gXY, gXY)
+          |}""".stripMargin
+      )).build().convertToInputWith("f9e5ce5aa0d95f5d54a7bc89c46730d9662397067250aa18a0039631c0f5b809", 0)
+
+      val txB = ctx.newTxBuilder()
+      val output = txB.outBoxBuilder().value(10000000).contract(ctx.compileContract(ConstantsBuilder.empty(),"{sigmaProp(true)}")).build()
+      val inputs = new java.util.ArrayList[InputBox]()
+      inputs.add(input)
+
+      val tree = "100207036ba5cfbc03ea2471fdf02737f64dbcd58c34461a7ec1e586dcd713dacbf89a120400d805d601db6a01ddd6027300d603b2a5730100d604e4c672030407d605e4c672030507eb02ce7201720272047205ce7201720472027205"
```
aslesarenko, line 44: Please add a comment with the original code of the tree.

common/src/main/java/org/ergoplatform/appkit/AppkitProvingInterpreter.scala
```diff
@@ -42,9 +43,10 @@ class AppkitProvingInterpreter(
   import Helpers._
 
   val secrets: Seq[SigmaProtocolPrivateInput[_ <: SigmaProtocol[_], _ <: SigmaProtocolCommonInput[_]]] = {
-    val dlogs = JListToIndexedSeq(identityIso[ExtendedSecretKey]).to(secretKeys).map(_.key)
-    val dhts = JListToIndexedSeq(identityIso[DiffieHellmanTupleProverInput]).to(dhtInputs)
-    dlogs ++ dhts
+    val dlogs: IndexedSeq[DLogProverInput] = try JListToIndexedSeq(identityIso[ExtendedSecretKey]).to(secretKeys).map(_.key) catch {case _:Exception =>  IndexedSeq()}
```
aslesarenko, line 46: Why this `try` is necessary?

lib-api/src/main/java/org/ergoplatform/appkit/ErgoProverBuilder.java
```diff
@@ -31,28 +33,26 @@
...
-
-    /**
-     * Add DHT prover input using this prover's secret as belonging to second party (i.e. Bob).
+     *   NOTE: We can swap x, y and obtain another tuple (g, u, h, v) with secret y (and unknown x).
+     *
+     * @param g {@Link GroupElement} instance defining g
+     * @param h {@Link GroupElement} instance defining h
+     * @param u {@Link GroupElement} instance defining u
+     * @param v {@Link GroupElement} instance defining v
+     * @param x {@Link BigInteger} instance defining x
```
aslesarenko, line 47: Could you provide more descriptive (actually useful) info. Otherwise it doesn't describe anything, which is not already known from the signature itself. Suppose I don't know what g, h, u, v means, can you guide me towards understanding of what they mean and where to look for.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ErgoProverBuilderImpl.java
```diff
@@ -52,31 +51,23 @@ public ErgoProverBuilder withSecretStorage(SecretStorage storage) {
...
-     */
     @Override
-    public ErgoProverBuilder withSecondDHTSecret(Address otherParty) {
-        checkState(_secondSecret == null, "Second secret already defined.");
-        BigInteger y = _masterKey.key().w();                         // Bob's secret key
-        SecP256K1Point g_y = _masterKey.key().publicImage().value(); // Bob's public key
-        SecP256K1Point g_x = otherParty.getPublicKey().value();
-        SecP256K1Point g_xy = CryptoConstants.dlogGroup().exponentiate(g_x, y);
-        _secondSecret = JavaHelpers.createDHTProverInput(g_x, g_y, y, g_xy);
+    public ErgoProverBuilder withDLogSecret(BigInteger x) {
```
aslesarenko, line 69: Please add ScalaDoc to the method.

appkit/src/test/scala/org/ergoplatform/appkit/DHTProverSpec.scala
scalahub: It can be either `gX.exp(y)` or `gY.exp(x)` as the operation is commutative, so should not matter.

appkit/src/test/scala/org/ergoplatform/appkit/DHTProverSpec.scala
scalahub: This was the ergo tree of some randomly selected box from explorer. I've added the box Id in a comment.

common/src/main/java/org/ergoplatform/appkit/AppkitProvingInterpreter.scala
scalahub: Otherwise `DHTProverSpec` throws a null pointer exception. See below

`java.lang.NullPointerException was thrown.
java.lang.NullPointerException
	at org.ergoplatform.appkit.AppkitProvingInterpreter.$anonfun$secrets$1(AppkitProvingInterpreter.scala:46)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
	at scala.collection.Iterator.foreach(Iterator.scala:941)
	at scala.collection.Iterator.foreach$(Iterator.scala:941)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableLike.map(TraversableLike.scala:238)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at org.ergoplatform.appkit.AppkitProvingInterpreter.<init>(AppkitProvingInterpreter.scala:46)
	at org.ergoplatform.appkit.impl.ErgoProverBuilderImpl.build(ErgoProverBuilderImpl.java:138)
	at org.ergoplatform.appkit.DHTProverSpec.$anonfun$new$2(DHTProverSpec.scala:50)
	at org.ergoplatform.appkit.FileMockedErgoClient.execute(FileMockedErgoClient.java:68)
	at org.ergoplatform.appkit.DHTProverSpec.$anonfun$new$1(DHTProverSpec.scala:25)
`

Line 46 is 
`val dlogs: IndexedSeq[DLogProverInput] = JListToIndexedSeq(identityIso[ExtendedSecretKey]).to(secretKeys).map(_.key)`

appkit/src/test/scala/org/ergoplatform/appkit/DHTProverSpec.scala
aslesarenko: Well, strictly speaking it matters, because it triggered my attention, high enough to write the question. Either fix the order or add a comment.

scalahub: Fixed, also removed ScalaDoc from impl class and put in abstract class.

common/src/main/java/org/ergoplatform/appkit/AppkitProvingInterpreter.scala
aslesarenko: Yes, but why NPE then? What is the root cause? It is probably some bug or misuse.

scalahub: Will investigate and update.

scalahub: Added a null check for _masterKey that was causing this. Essentially, when the prover builder was initialized without master secret, it was giving this error.

scalahub: Made the use of key like the other parameters:

        ArrayList<ExtendedSecretKey> keys = new ArrayList<>();
        if (_masterKey != null) keys.add(_masterKey);

scalahub: NPE occurs because the wallet is not initialized when the prover is created. The above change sets the initial wallet keys similar to how additional keys are set (i.e., with a null check). 
Ideally wallet need not be initialized in many scripts, for example `sigmaProp(1 < 2)`

lib-api/src/main/java/org/ergoplatform/appkit/ErgoProverBuilder.java
scalahub: Could you check latest commit and see if it is resolved?

PR #71, opened 2021-03-31
Allow set creationHeight in OutBoxBuilder

### LINKED ISSUE #69: Allow setting custom creationHeight in OutBoxBuilder
When creating custom output boxes using OutBoxBuilder, we should be able to specify the creationHeight.

aslesarenko:
Resolves #69


lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  override def creationHeight(height: Int): OutBoxBuilder = {
+    _creationHeightOpt = Some(height)
+    this
+  }
+
+  override def build: OutBox = {
+    checkState(_contract != null, "Contract is not defined": Any)
+    val tree = _contract.getErgoTree
+    // TODO pass user specified creationHeight
```
scalahub, line 64: This comment should be removed

scalahub (changes requested):

scalahub:
One small comment; otherwise ok

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
aslesarenko: fixed

PR #75, opened 2021-04-20
Correct cost accumulation when tx is signed

### LINKED ISSUE #74: Getting "Cost of transaction ErgoLikeTransactionTemplate( ... ) exceeds limit [...]" error
Getting error `Cost of transaction ErgoLikeTransactionTemplate( ... ) exceeds limit 5062531`

In appkit test node responses, I've set the `maxBlockCost` to be 5062531 as per the current mainnet (ref @kushti )

Full error here https://gist.github.com/scalahub/67a8293010428ea2bbfa2085690116fc

Here is the test that generates the error
https://github.com/scalahub/AppkitIssue/blob/main/src/test/scala/AppkitIssue.scala

The same transaction (plus few additional inputs) when created via the ergo node works fine: 
https://explorer.ergoplatform.com/en/transactions/eab3020a702165b1328f720df99cee051b5c282b52fde563c291f97113bad1f3
We can verify that the addresses in the above tx and the test (AppkitIssue.scala) are the same

For reference, the scripts of the addresses correspond the code here:

ErgoScript: https://github.com/scalahub/OraclePool/blob/main/src/main/scala/oraclepool/v1a/Contracts.scala
Addresses: https://github.com/scalahub/OraclePool/blob/main/src/test/scala/oraclepool/v1a/AddressSpec.scala

aslesarenko:
Fixes #74


common/src/main/java/org/ergoplatform/appkit/AppkitProvingInterpreter.scala
```diff
@@ -53,27 +54,54 @@ class AppkitProvingInterpreter(
...
+
+    val (outAssets, outAssetsNum) = JavaHelpers.extractAssets(unsignedTx.outputCandidates)
+    val (inAssets, inAssetsNum) = JavaHelpers.extractAssets(boxesToSpend.map(_.box))
+
+    val tokenAccessCost = params.tokenAccessCost
+    val totalAssetsAccessCost =
+      Math.addExact(
+        Math.multiplyExact(Math.addExact(outAssetsNum, inAssetsNum), tokenAccessCost),
+        Math.multiplyExact(Math.addExact(inAssets.size, outAssets.size), tokenAccessCost))
+    currentCost = addCost(currentCost, totalAssetsAccessCost, maxCost, "")
```
scalahub, line 101: Wouldn't a message be better instead of empty string?

scalahub (approved):
One minor comment (better to use non-empty error message for token cost computation), otherwise looks good

common/src/main/java/org/ergoplatform/appkit/AppkitProvingInterpreter.scala
aslesarenko: Yeah, added a message.

PR #78, opened 2021-05-16
Allow transactions with user-defined fee output

### LINKED ISSUE #77: Allow "exact" transactions
From discord.
scalahub — 05/10/2021
Can you modify appkit to accept zero fee when creating tx.. There is a situation where I need to add fee output manually and not via appkit.
Basically another method that allows user to specify the exact tx without appkit adding anything extra.

aslesarenko:
Fixes #77


lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
```diff
@@ -0,0 +1,133 @@
...
+  private var _ph: PreHeaderImpl = null
+
+  override def preHeader(ph: PreHeader): UnsignedTransactionBuilder = {
+    require(_ph == null, "PreHeader is already specified")
+    _ph = ph.asInstanceOf[PreHeaderImpl]
+    this
+  }
+
+  override def boxesToSpend(inputBoxes: List[InputBox]): UnsignedTransactionBuilder = {
+    val items = inputBoxes
```
scalahub, line 39: Wouldn't it be better to have null check for _inputBoxes as we do for preHeader?

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
```diff
@@ -0,0 +1,133 @@
...
+      .map(box => JavaHelpers.createUnsignedInput(box.getId.getBytes))
+    _inputs.addAll(items)
+    _inputBoxes = inputBoxes.map(b => b.asInstanceOf[InputBoxImpl])
```
scalahub, line 42: There seems to be some inconsistency in the way two collections are handled.
addAll appends the items to _inputs, while the next line overwrites _inputBoxes. Is this by design?
Same comment applies to other methods too where two collections are updated

appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
```diff
@@ -89,4 +89,33 @@ class TxBuilderSpec extends PropSpec with Matchers
...
+        .value(30000000)
+        .contract(truePropContract(ctx))
+        .build()
+        .convertToInputWith(mockTxId, 0)
+      val output = txB.outBoxBuilder()
+        .value(29000000)
+        .contract(truePropContract(ctx)).build()
+      val feeOut = txB.outBoxBuilder()
+        .value(1000000)
+        .contract(ctx.newContract(ErgoScriptPredef.feeProposition()))
```
scalahub, line 107: Can we have another test where some other proposition is added?

scalahub (changes requested):

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
aslesarenko: fixed

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
aslesarenko: fixed

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
```diff
@@ -65,36 +68,46 @@ class UnsignedTransactionBuilderImpl(val _ctx: BlockchainContextImpl) extends Un
...
+    val dataInputBoxes = _dataInputBoxes
+       .getOrElse(new util.ArrayList[InputBoxImpl]())
+       .map(b => b.getErgoBox)
     val dataInputs = JavaHelpers.toIndexedSeq(_dataInputs)
     require(_feeAmount.isEmpty || _feeAmount.get >= MinFee,
       s"When fee amount is defined it should be >= $MinFee, got ${_feeAmount.get}")
-    require(_changeAddress != null, "Change address is not defined")
-    val outputCandidates = JavaHelpers.toIndexedSeq(_outputCandidates)
+    require(_changeAddress.isDefined, "Change address is not defined")
+    val outputCandidates = JavaHelpers.toIndexedSeq(_outputCandidates.get)
```
scalahub, line 102: Would be good to remove get as its considered anti-pattern

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
```diff
@@ -65,36 +68,46 @@ class UnsignedTransactionBuilderImpl(val _ctx: BlockchainContextImpl) extends Un
...
     val inputBoxes = JavaHelpers.toIndexedSeq(boxesToSpend.map(eb => eb.box))
-    val burnTokens = JavaHelpers.createTokensMap(Iso.isoJListErgoTokenToMapPair.to(_tokensToBurn))
+    val burnTokens = JavaHelpers.createTokensMap(
+      Iso.isoJListErgoTokenToMapPair.to(_tokensToBurn.getOrElse(new ArrayList[ErgoToken]))
+    )
     val tx = TransactionBuilder.buildUnsignedTx(
       inputs = inputBoxes, dataInputs = dataInputs, outputCandidates = outputCandidates,
       currentHeight = _ctx.getHeight, createFeeOutput = _feeAmount,
-      changeAddress = _changeAddress, minChangeValue = MinChangeValue,
+      changeAddress = _changeAddress.get, minChangeValue = MinChangeValue,
```
scalahub, line 110: Same as earlier comment about get

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
```diff
@@ -120,14 +136,13 @@ class UnsignedTransactionBuilderImpl(val _ctx: BlockchainContextImpl) extends Un
...
-    if (_ph == null) _ctx.getPreHeader else _ph
+  override def getPreHeader: PreHeader = _ph.getOrElse(_ctx.getPreHeader)
 
   override def outBoxBuilder = new OutBoxBuilderImpl(this)
 
   override def getNetworkType: NetworkType = _ctx.getNetworkType
 
   override def getInputBoxes: List[InputBox] =
-    _inputBoxes.stream.collect(Collectors.toList[InputBox])
+    _inputBoxes.get.stream.collect(Collectors.toList[InputBox])
```
scalahub, line 146: Ref. earlier comment

scalahub (approved):
Few minor comments (use of "get" in Option type) that can be optionally addressed. Although its checked earlier using "isDefined", use of "get" is still considered anti-pattern.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
aslesarenko: fixed

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
aslesarenko: fixed

lib-impl/src/main/java/org/ergoplatform/appkit/impl/UnsignedTransactionBuilderImpl.scala
aslesarenko: fixed

PR #83, opened 2021-06-08
Update api client classes (Node and Explorer)

aslesarenko:
- [x] Ergo Node API classes 
- [x] Explorer API classes
- [x] explorerUrl parameter for RestApiErgoClient


MrStahlfelge (commented):
The API I tested with - `getApiV1AddressesP1BalanceTotal` - works. But only on Mainnet.
However, on Testnet it just returns a 404 ([test](https://api-testnet.ergoplatform.com/api/v1/addresses/3Wwxnaem5ojTfp91qfLw3Y4Sr7ZWVcLPvYSzTsZ4LKGcoxujbxd3/balance/total)). That is not appkit's fault, but it looks like Testnet explorer is not configured correctly. 

A suggestion for appkit I have is to have the API URLs from [here](https://github.com/ergoplatform/ergo-appkit/blob/197a7c724cb7c36a9227a81cfa8626c6f912ffd1/appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java#L32) moved to a public constant that can be used by clients.

appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
```diff
@@ -14,26 +14,38 @@
     private final String _nodeUrl;
     private final NetworkType _networkType;
     private final ApiClient _client;
+    private final String _explorerUrl;
     private final ExplorerApiClient _explorer;
 
+    public static String defaultMainnetExplorerUrl = "https://api.ergoplatform.com";
```
MrStahlfelge, line 20: Thank you! I think these two constants could be final as well?

aslesarenko: thanks, fixed

PR #88, opened 2021-06-12
Support EIP-3 addresses

aslesarenko:


MrStahlfelge (commented):
Thank you, tested with the snapshot and this gives the same addresses as Yoroi uses.

What I think we need as well are new methods for `ErgoProverBuilder` to construct the secrets for EIP-3 from the seed?

MrStahlfelge (approved):
Tested creating addresses and sending payments, works as expected. Thank you.

PR #90, opened 2021-06-21
Implemented mnemonicToSeed based on bouncycastle 

### LINKED ISSUE #82: java.security.NoSuchAlgorithmException: SecretKeyFactory PBKDF2WithHmacSHA512 implementation not found
The following error is thrown on older Android devices (< 8.0) when Address.fromMnemonic is called:

```
java.security.NoSuchAlgorithmException: SecretKeyFactory PBKDF2WithHmacSHA512 implementation not found
        at org.apache.harmony.security.fortress.Engine.notFound(Engine.java:190)
        at org.apache.harmony.security.fortress.Engine.getInstance(Engine.java:139)
        at javax.crypto.SecretKeyFactory.getInstance(SecretKeyFactory.java:108)
        at org.ergoplatform.wallet.mnemonic.Mnemonic$.toSeed(Mnemonic.scala:77)
        at org.ergoplatform.appkit.JavaHelpers$.seedToMasterKey(JavaHelpers.scala:354)
        at org.ergoplatform.appkit.JavaHelpers.seedToMasterKey(JavaHelpers.scala)
        at org.ergoplatform.appkit.Address.fromMnemonic(Address.java:135)
```
Algorithm is not included in Java 7. As you already ship with Bouncy castle, it should be easy to switch to its implementation that is available on all language levels. See here: https://stackoverflow.com/a/22621191/7487013

aslesarenko:
closes #82


MrStahlfelge (approved):
This works like a charm.

You can consider to only use the new method only as a fallback in case that `NoSuchAlgorithmException` is actually raised, but I don't know if there are ups or downs when using the built-in method.

PR #93, opened 2021-07-14
Work without explorer

### LINKED ISSUE #92: set parameter explorerUrl as an optional field
https://github.com/ergoplatform/ergo-appkit/blob/273d99d82a6479827fc32eeacc133ddc9e02a4a8/appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java#L33

According to the discussion in discord, I think it is better if explorerUrl to be optional so that Appkit method can work with Node and Explorer `switch case defaultMainnetExplorerUrl or defaultTestnetExplorerUrl` as long as specific Explorer is actually not needed.

aslesarenko:
Closes #92

### DIFF
--- a/appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
+++ b/appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
@@ -66,11 +66,25 @@ public class RestApiErgoClient implements ErgoClient {
     }
 
     /**
+     * Creates a new {@link ErgoClient} instance in the `node-only` mode, i.e. connected
+     * to a given node of the given network type and not connected to explorer.
+     *
+     * @param nodeUrl     http url to Ergo node REST API endpoint of the form `https://host:port/`
+     * @param networkType type of network (mainnet, testnet) the Ergo node is part of
+     * @param apiKey      api key to authenticate this client
+     * @return a new instance of {@link ErgoClient} connected to a given node
+     */
+    public static ErgoClient createWithoutExplorer(String nodeUrl, NetworkType networkType, String apiKey) {
+        return new RestApiErgoClient(nodeUrl, networkType, apiKey, null);
+    }
+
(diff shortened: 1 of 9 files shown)


MrStahlfelge (commented):
Interesting: 

I was not aware that explorerUrl was already optional and, if not given, a default value is used. Because of this, I understood the request on Discord in a way so that it actually is not needed to give the explorer url, more precisley to add a new method `public static ErgoClient create(String nodeUrl, NetworkType networkType, String apiKey)`.

If explorer url is not needed in all cases, this PR makes sense. However, JavaDoc of `create` method should be changed, too (or constructor should be made public and create method removed)

MrStahlfelge:
@zargarzadehm can you tell us if your original request was because you also did not know that `null` could be given, or if you actually wanted to have no ExplorerApiClient initialization?

zargarzadehm:
I think the previous code was ok, just sometimes it is necessary to use a special explorer (in my case I have a separate internal network for the development environment), I think by default there should be no need to enter the exploreUrl (The same Mainnet and Testnet are sufficient), so I think the following change is enough.
```
public static ErgoClient create(String nodeUrl, NetworkType networkType, String apiKey, String explorerUrl = "" `or null` )
```

But the current implemented logic is also a good logic.

aslesarenko:
@MrStahlfelge @zargarzadehm the conditional initialization of Explorer client may be useful in situation (hypothetical) when local node is accessed, but remote explorer is not available. It will still be possible to build a transaction and send it via Node API alone.

MrStahlfelge:
Yep, I understand that the change is useful. I think just a clarification on Javadoc `RestErgoApiClient.create` needed (or removing the need to call the static create method as an alternative).

zargarzadehm (commented):
Looks good to me

MrStahlfelge (approved):
Perfect now

PR #95, opened 2021-07-18
fix bug set explorer client

zargarzadehm:


MrStahlfelge (approved):
Thank you, we really missed it!

PR #97, opened 2021-07-20
Collection of CoveringBoxes using chunked requests

### LINKED ISSUE #96: limit on return unspent boxes on explorer
https://github.com/ergoplatform/ergo-appkit/blob/c0e0db3e10e6903c22e9bc8f18e99a2e714e011a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/ExplorerFacade.java#L31

Now explorerApi `V1`  just return `20` unspent boxes, for getting more unspent box should be used query param `limit` but in scala side, there is not this feature.
Other APIs may have this limitation from Explorer.

aslesarenko:
In this PR:
- [x] Closes #96 
- [x] add `offset`, `limit` parameters to `getUnspentBoxesFor`
- [x] add new method `getCoveringBoxesFor`
- [x] more missing JavaDocs

### DIFF
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
@@ -77,7 +77,7 @@ public interface BlockchainContext {
 
     ErgoContract compileContract(Constants constants, String ergoScript);
 
-    /** Default size of the chuck (aka page size) used in API requests. */
+    /** Default size of the chunk (aka page size) used in API requests. */
     int DEFAULT_LIMIT_FOR_API = 20;
 
     /**
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
@@ -77,9 +77,26 @@ public interface BlockchainContext {
 
     ErgoContract compileContract(Constants constants, String ergoScript);
 
+    int DEFAULT_LIMIT_FOR_API = 20;
+
     /**
-     * Get unspent boxes owned by the given address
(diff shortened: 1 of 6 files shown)


zargarzadehm:
@aslesarenko 
In some usage of `getUnspentBoxesFor` we want to get all boxes and filters them according to specific tokens or other conditions (registers, ....).
Given this implementation, if we needed to get all boxes, do you have any method other than to progress blindly using the `limit` and `offset`?
I think if a method is implemented that can return the value of key `total` is appropriate.

aslesarenko:
> Given this implementation, if we needed to get all boxes, do you have any method other than to progress blindly using the `limit` and `offset`?

Yes, you can use getCoveringBoxesFor method with very very large amountToSpend (like 100000000 * 1000000000 which is greater than total amount of ergs emission). The method will retrieve all the boxes, page by page, for the given address)

zargarzadehm (approved):

lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
```diff
@@ -77,9 +77,29 @@
 
     ErgoContract compileContract(Constants constants, String ergoScript);
 
+    /** Default size of the chuck (aka page size) used in API requests. */
```
MrStahlfelge, line 80: Chunk instead of chuck?

MrStahlfelge (approved):
This looks like a good change to me! I did not test it though (v1 testnet problem).
I might have found a typo, please check.

lib-api/src/main/java/org/ergoplatform/appkit/BlockchainContext.java
aslesarenko: thanks, fixed

PR #103, opened 2021-08-13
Sending ERG: Retry when NotEnoughCoinsForChangeBoxesError occurs

MrStahlfelge:
Problem: When sending ERG, it could happen that an Inbox with an asset is selected. 

No problem in general, the asset is returned with a change box.

But when the amount of the selected Inbox can only satisfy the amount to send to the recipient, the NotEnoughCoinsForChangeBoxesError occurs. The user can't work around it, although the workaround is easy (when enough balance given): Just select another inbox. 

This approach makes a single retry to find more inboxes when the NotEnoughCoinsForChangeBoxesError occurs.

If you don't like this approach, I am open for other suggestions as well. Please also check the Scala file as I am not sure if the instance check is well done.

### DIFF
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BoxSelectorsJavaHelpers.scala
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BoxSelectorsJavaHelpers.scala
@@ -3,12 +3,14 @@ package org.ergoplatform.appkit
 import scala.collection.mutable
 import org.ergoplatform.wallet.boxes.DefaultBoxSelector
 import java.util.{List => JList, Map => JMap}
+
 import org.ergoplatform.appkit.JavaHelpers._
 import org.ergoplatform.appkit.Iso._
 import org.ergoplatform.ErgoBox.TokenId
 import org.ergoplatform.ErgoBoxAssets
 import org.ergoplatform.ErgoBoxAssetsHolder
-import scorex.util.{bytesToId, ModifierId}
+import org.ergoplatform.wallet.boxes.DefaultBoxSelector.NotEnoughCoinsForChangeBoxesError
+import scorex.util.{ModifierId, bytesToId}
 
 
 object BoxSelectorsJavaHelpers {
@@ -26,9 +28,13 @@ object BoxSelectorsJavaHelpers {
       .map(InputBoxWrapper.apply).toIterator
     val targetAssets = tokensToSpend.convertTo[mutable.LinkedHashMap[ModifierId, Long]].toMap
     val foundBoxes: IndexedSeq[InputBox] = DefaultBoxSelector.select(inputBoxes, amountToSpend, targetAssets) match {
-      case Left(err) => 
-        throw new RuntimeException(
-          s"Not enough funds in boxes to pay $amountToSpend nanoERGs, \ntokens: $tokensToSpend, \nreason: $err")
+      case Left(err) =>
+        if (err.isInstanceOf[NotEnoughCoinsForChangeBoxesError]) {
+          throw new InputBoxesSelectionException.NotEnoughCoinsForChangeException(err.message)
+        } else {
+          throw new InputBoxesSelectionException(
+            s"Not enough funds in boxes to pay $amountToSpend nanoERGs, \ntokens: $tokensToSpend, \nreason: $err")
+        }
       case Right(v) => v.boxes.map(_.inputBox).toIndexedSeq
     }
     foundBoxes.convertTo[JList[InputBox]]
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.iml
 *.toc
 *.aux
 *.log
--- a/common/src/main/java/org/ergoplatform/appkit/InputBoxesSelectionException.java
+++ b/common/src/main/java/org/ergoplatform/appkit/InputBoxesSelectionException.java
@@ -1,0 +1,19 @@
+package org.ergoplatform.appkit;
+
+public class InputBoxesSelectionException extends RuntimeException {
+
+    public InputBoxesSelectionException(String message) {
+        super(message);
+    }
+
+    /**
+     * Thrown when a change box is needed, but ERG amount in all inboxes is not enough to create the
+     * change box
+     */
+    public static class NotEnoughCoinsForChangeException extends InputBoxesSelectionException {
+
+        public NotEnoughCoinsForChangeException(String message) {
+            super(message);
+        }
+    }
+}
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
@@ -149,7 +149,21 @@ public class BoxOperations {
         } else {
             senders.add(senderProver.getAddress());
         }
-        List<InputBox> boxesToSpend = loadTop(ctx, senders, amountToSend + MinFee, tokensToSpend);
+
+        List<InputBox> boxesToSpend;
+        try {
+            boxesToSpend = loadTop(ctx, senders, amountToSend + MinFee, tokensToSpend);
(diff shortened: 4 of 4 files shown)


MrStahlfelge:
Converted to draft because when tested, it did not work - when building the UnsignedTransaction, IllegalArgumentException(   "failed to calculate change for outputTotal"...) is raised from TransactionBuilder.scala. I am not sure if this is a bug or intended, so I am waiting for input by @aslesarenko 

All in all, this is an edge case and a work around is just raising the ERG amount to be send. But sometimes this is not possible for the user.

lib-api/src/main/java/org/ergoplatform/appkit/BoxSelectorsJavaHelpers.scala
```diff
@@ -26,9 +28,13 @@ object BoxSelectorsJavaHelpers {
...
-      case Left(err) => 
-        throw new RuntimeException(
-          s"Not enough funds in boxes to pay $amountToSpend nanoERGs, \ntokens: $tokensToSpend, \nreason: $err")
+      case Left(err) =>
+        if (err.isInstanceOf[NotEnoughCoinsForChangeBoxesError]) {
+          throw new InputBoxesSelectionException.NotEnoughCoinsForChangeException(err.message)
+        } else {
+          throw new InputBoxesSelectionException(
+            s"Not enough funds in boxes to pay $amountToSpend nanoERGs, \ntokens: $tokensToSpend, \nreason: $err")
+        }
```
aslesarenko, line 37: @MrStahlfelge In Scala you can use type test embedded in the pattern like
```
case Left(err: NotEnoughCoinsForChangeBoxesError) =>
    throw new NotEnoughCoinsForChangeException(err.message)
case Left(err) =>
    throw new InputBoxesSelectionException(...)
```

lib-api/src/main/java/org/ergoplatform/appkit/BoxSelectorsJavaHelpers.scala
```diff
@@ -26,9 +28,13 @@ object BoxSelectorsJavaHelpers {
       .map(InputBoxWrapper.apply).toIterator
     val targetAssets = tokensToSpend.convertTo[mutable.LinkedHashMap[ModifierId, Long]].toMap
     val foundBoxes: IndexedSeq[InputBox] = DefaultBoxSelector.select(inputBoxes, amountToSpend, targetAssets) match {
-      case Left(err) => 
-        throw new RuntimeException(
-          s"Not enough funds in boxes to pay $amountToSpend nanoERGs, \ntokens: $tokensToSpend, \nreason: $err")
+      case Left(err) =>
+        if (err.isInstanceOf[NotEnoughCoinsForChangeBoxesError]) {
+          throw new InputBoxesSelectionException.NotEnoughCoinsForChangeException(err.message)
```
aslesarenko, line 33: Why repackage original exception in a new one?

aslesarenko (commented):
LGTM, please see comments.

aslesarenko:
> IllegalArgumentException( "failed to calculate change for outputTotal"...)

what is in the original BoxSelectorError? (the `err` value https://github.com/ergoplatform/ergo/blob/323dd2ce3149db3758df03f7cf68242bae53bc85/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala#L105)

lib-api/src/main/java/org/ergoplatform/appkit/BoxSelectorsJavaHelpers.scala
MrStahlfelge: Idea is that we have different Java exceptions for all three error types that can occur so that clients can determine which error was thrown by the Scala code. The Java Exceptions for NotEnoughErgsException and NotEnoughTokensException should also provide fields like the ones added to the Scala Errors in ergoplatform/ergo#1423 - I did not add them here though, because it has nothing to do with the problem this PR tries to solve.

MrStahlfelge:
> > IllegalArgumentException( "failed to calculate change for outputTotal"...)
> 
> what is in the original BoxSelectorError? (the `err` value https://github.com/ergoplatform/ergo/blob/323dd2ce3149db3758df03f7cf68242bae53bc85/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala#L105)

This is the full error with stacktrace - looks like the second Inbox is ignored?

```
java.lang.IllegalArgumentException: failed to calculate change for outputTotal: 2000000, 
tokens: Map(f9ff04b93ff67ec2de04b05675e41a38c0faf409d499fd344c6d34e5e9a4bedd -> 1), 
burnTokens: Map(), 
inputs: Vector(ErgoBox(01d285c0e6f9b9022145acf613229aea37b4f60cf91a9cc11e205cb7e2cdd9c2,2000000,ErgoTree(0,WrappedArray(),Right(ConstantNode(SigmaProp(ProveDlog(ECPoint(2de726,af561f,...))),SSigmaProp)),80,[B@e6bd094,Some(false)),tokens: (Coll(f9ff04b93ff67ec2de04b05675e41a38c0faf409d499fd344c6d34e5e9a4bedd:1,cda562fd3a4dca91cbc2075a911a3a5b11b7ed810097b1ac5a54b762a877b062:100000000000,514083a170fc734071c07748ff444940606654317bd516865120ed702952ab1b:100000000000,96d65b193899c154d75bc3a55b2d3c2311f09f02b9d6ed2dac913cb10c383080:1000000000)), 8c77268419ec66de08f84e325cef6669452912755a365a818469c40f6b1ee79f, 2, Map(), 51952), ErgoBox(d51bc2037e4bb6545d7bb93fdb1368185bfa539cfb6250abd95ad565914de1b9,100000000000,ErgoTree(0,WrappedArray(),Right(ConstantNode(SigmaProp(ProveDlog(ECPoint(2de726,af561f,...))),SSigmaProp)),80,[B@3ed173d,Some(false)),tokens: (Coll()), 7851af9cf8406f1639d3c2c589ff408af95c43ced731cb6f8a77b55e6f2b5a06, 0, Map(), 51956)), 
reason: NotEnoughCoinsForChangeBoxesError(Not enough ERG 0 to create 1 change boxes, 
for Stream(Map(cda562fd3a4dca91cbc2075a911a3a5b11b7ed810097b1ac5a54b762a877b062 -> 100000000000, 514083a170fc734071c07748ff444940606654317bd516865120ed702952ab1b -> 100000000000, 96d65b193899c154d75bc3a55b2d3c2311f09f02b9d6ed2dac913cb10c383080 -> 1000000000)))
	at org.ergoplatform.wallet.transactions.TransactionBuilder$$anonfun$buildUnsignedTx$1.apply(TransactionBuilder.scala:105)
	at org.ergoplatform.wallet.transactions.TransactionBuilder$$anonfun$buildUnsignedTx$1.apply(TransactionBuilder.scala:77)
	at scala.util.Try$.apply(Try.scala:192)
	at org.ergoplatform.wallet.transactions.TransactionBuilder$.buildUnsignedTx(TransactionBuilder.scala:77)
	at org.ergoplatform.appkit.impl.UnsignedTransactionBuilderImpl.build(UnsignedTransactionBuilderImpl.scala:119)
	at org.ergoplatform.appkit.BoxOperations.putToContractTx(BoxOperations.java:182)
```

aslesarenko:
That is right, for totalAmount of 2000000 the first box is already enough. So the second in not selected. 
Due to MaxTokens=255 per box limit the logic for creating change boxes is not trivial. See this method https://github.com/ergoplatform/ergo/blob/eba65deda77c4da81ecc8c6502ba681f30fa3807/ergo-wallet/src/main/scala/org/ergoplatform/wallet/boxes/DefaultBoxSelector.scala#L88 where you can find the error.

MrStahlfelge:
Some of this already merged, and we can do better than just retry. Will do a new approach.

PR #107, opened 2021-09-29
Construct address from ErgoTree: support p2pk addresses

MrStahlfelge:
`Address.fromErgotree()` did only work for p2s addresses correctly. This changes it to support p2pk addresses as well.


aslesarenko:
Looks good.

aslesarenko (approved):

PR #114, opened 2021-12-06
Pass-through original DefaultBoxSelector error information

MrStahlfelge:
For MrStahlfelge/ergo-wallet-android#25

This enables client to access the original error information returned from DefaultBoxSelector by repacking it into objects that Java can handle.

Not sure if my conversion into HashMap is the way to go, if there is a  better way please let me know.

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/InputBoxesSelectionException.java
+++ b/common/src/main/java/org/ergoplatform/appkit/InputBoxesSelectionException.java
@@ -1,0 +1,45 @@
+package org.ergoplatform.appkit;
+
+import java.util.Map;
+
+public class InputBoxesSelectionException extends RuntimeException {
+
+    public InputBoxesSelectionException(String message) {
+        super(message);
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/ergoplatform/appkit/InputBoxesSelectionException.java
```diff
@@ -0,0 +1,45 @@
+package org.ergoplatform.appkit;
+
+import java.util.Map;
+
+public class InputBoxesSelectionException extends RuntimeException {
```
aslesarenko, line 5: What is the motivation to introduce this wrappers around ergo-wallet exceptions?

aslesarenko (approved):
LGTM, please see comments

common/src/main/java/org/ergoplatform/appkit/InputBoxesSelectionException.java
MrStahlfelge: It enables clients to access the original error information and beautify it for the user. Most important is to show ERG amounts instead of nanoergs for NotEnoughErgsException, and additionally showing a list instead of a serialized map for tokens.

PR #118, opened 2021-12-30
Add openapi config and input for node api

MrStahlfelge:
This adds openapi config and input file for updating the Node API.

On this branch, you can run `sbt javaClientGenerated/openApiGenerate` to update the Node API files. However, the Node config is either corrupted or something else is wrong, the files won't compile at least due to a missing Int64 type.

Explorer API could be added the same way, but as I understand the sbt plugin this would need a new module. But we should bring this in a useable state first.


MrStahlfelge:
When this is resolved, we can add an abstract api client layer (that could be configured what to use explorer and node for). ExplorerAndPoolUnspentBoxesLoader will then be able to use this abstract layer.
GraalVM face implementations will be removed.

MrStahlfelge:
Closed, switched to manual refinement of explorer APIs when needed

PR #119, opened 2022-01-02
Added option proxy connection to clients

anon2020s:


appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
```diff
@@ -33,14 +34,23 @@
      *                    form `https://host[:port]` where port is optional.
      *                    If `null` or empty string passed then the Explorer client is not
      *                    initialized and the client works in the `node only` mode.
+     * @param proxy       Requests are passed through this proxy (null it's ok).
```
MrStahlfelge, line 37: "if non-null" instead of "null it's ok"?

appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
```diff
@@ -93,7 +103,26 @@ public static ErgoClient createWithoutExplorer(String nodeUrl, NetworkType netwo
...
+     *
+     * @param nodeUrl     http url to Ergo node REST API endpoint of the form
+     * `https://host:port/`
+     * @param networkType type of network (mainnet, testnet) the Ergo node is part of
+     * @param apiKey      api key to authenticate this client
+     * @param explorerUrl optional http url to Explorer REST API endpoint of the form
+     *                    `https://host:port/`. If null or empty, then explorer connection
+     *                    is not initialized so that the resulting {@link ErgoClient} can
+     *                    work in `node-only` mode.
+     * @param proxy       Requests are passed through this proxy (null it's ok).
```
MrStahlfelge, line 121: Same as above.

Additionally, it should be stated that proxy is only used for Explorer API, not node. Maybe the param should also be renamed to "explorerApiProxy" because of this.

By the way, we could start using `@Nullable` annotations here as well. Made a quick check and it seems javax.annotation.Nullable is on the classpath

appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
anon2020s: Fixed!

appkit/src/main/java/org/ergoplatform/appkit/RestApiErgoClient.java
anon2020s: The mixer uses public nodes in the network, so a proxy is required to communicate with the nodes and the explorer.

MrStahlfelge: Sorry, did not see that proxy is used for node api as well.

MrStahlfelge (approved):
This looks good to me. 

@aslesarenko do you know what's wrong with publishing the snapshot?

aslesarenko:
@MrStahlfelge switching of publishing seems to work

PR #100, opened 2021-08-06
Java7 compatibility

MrStahlfelge:
Follow-up to ScorexFoundation/sigmastate-interpreter#736

PR #120, opened 2022-01-07
Updated sigmastate-interpreter to 4.0.5, updated ergo-wallet to 4.0.20

MrStahlfelge:
Use latest release of ergo-wallet and sigmastate-interpreter

(Follow-up to #100)

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/SecretString.java
+++ b/common/src/main/java/org/ergoplatform/appkit/SecretString.java
@@ -93,4 +93,8 @@ public final class SecretString {
     public String toStringUnsecure() {
         return String.valueOf(_data);
     }
+
+    public org.ergoplatform.wallet.interface4j.SecretString toInterface4JSecretString() {
+        return org.ergoplatform.wallet.interface4j.SecretString.create(_data.clone());
+    }
 }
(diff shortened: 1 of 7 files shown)


MrStahlfelge:
We might consider moving to ergo-wallets SecretString, but it would be a breaking change because of the package change and some methods missing.

common/src/main/java/org/ergoplatform/appkit/SecretString.java
```diff
@@ -93,4 +93,8 @@ public boolean equals(Object obj) {
     public String toStringUnsecure() {
         return String.valueOf(_data);
     }
+
+    public org.ergoplatform.wallet.interface4j.SecretString toInterface4JSecretString() {
+        return org.ergoplatform.wallet.interface4j.SecretString.create(_data.clone());
+    }
```
aslesarenko, line 99: Hm, I think we don't need two implementations of SecretString. 
Any reason we cannot just use org.ergoplatform.wallet.interface4j.SecretString everywhere?

aslesarenko (changes requested):

common/src/main/java/org/ergoplatform/appkit/SecretString.java
MrStahlfelge: As noted, this would be a breaking change because of the package change and some methods missing. I don't know why the class wasn't adopted unchanged. If we do a breaking change, next version should be 4.1.x or 5.0.x and not 4.0.6. But I think after all this time, a non-breaking bugfix release is necessary.

aslesarenko (approved):

PR #124, opened 2022-01-26
BlockchainContextImpl.getUnspentBoxesFor: skip boxes unknown to node

MrStahlfelge:
`BlockchainContextImpl.getUnspentBoxesFor` returns a list of unspent boxes. However, when the node it connects to is not in full sync, certain box information can't be retrieved. This causes an NPE and no unspend box can be retrieved.

With this change, instead of throwing an NPE a null object is added to the list. This way, the applications can see that some information is missing (and not confuse a smaller chunk size with the end of the list), but can continue working with other available unspent boxes.

### DIFF
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
@@ -98,7 +98,11 @@ public class BlockchainContextImpl extends BlockchainContextBase {
         for (OutputInfo box : boxes) {
             String boxId = box.getBoxId();
             ErgoTransactionOutput boxInfo = ErgoNodeFacade.getBoxById(_retrofit, boxId);
-            returnList.add(new InputBoxImpl(this, boxInfo));
+            // can be null if node does not know about the box (yet)
+            // instead of throwing an error, we continue with the boxes actually known
+            if (boxInfo != null) {
+                returnList.add(new InputBoxImpl(this, boxInfo));
+            }
         }
 
         return returnList;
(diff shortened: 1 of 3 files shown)


lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextImpl.java
```diff
@@ -98,7 +98,14 @@ ApiClient getApiClient() {
...
             String boxId = box.getBoxId();
             ErgoTransactionOutput boxInfo = ErgoNodeFacade.getBoxById(_retrofit, boxId);
-            returnList.add(new InputBoxImpl(this, boxInfo));
+            // can be null if node does not know about the box (yet)
+            // instead of throwing an error, we continue with the boxes actually known
+            if (boxInfo != null) {
+                returnList.add(new InputBoxImpl(this, boxInfo));
+            } else {
+                // add a placeholder - otherwise we cannot determine if we are at the end of the list
+                returnList.add(null);
```
aslesarenko, line 107: I don't think this is a good idea. It is better to filter such boxes out. 
In order to detect the last chunk the `chunk.size() < DEFAULT_LIMIT_FOR_API` can be replaced with `chunk.size() == 0`.
This will require one additional request, but on the other hand will not rely on each chuck be of the DEFAULT_LIMIT_FOR_API size.

aslesarenko (changes requested):

aslesarenko (approved):

MrStahlfelge:
The tests run successful, but GHA actions don't check that they were done. Looks like a GitHub actions bug for me.

PR #128, opened 2022-02-05
Add methods to serialize and deserialize an extended public key according to bip-32

MrStahlfelge:
Also added method  to derive eip3 addresses from such a key

Needed for ergoplatform/ergo-wallet-app#32

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/Address.java
+++ b/common/src/main/java/org/ergoplatform/appkit/Address.java
@@ -173,13 +173,11 @@ public class Address {
 
         // Let's use "m/44'/429'/0'/0/index" path (this path is compliant with EIP-3 which
         // is BIP-44 for Ergo)
-        DerivationPath path = JavaHelpers.eip3DerivationWithLastIndex(index);
+        DerivationPath path = JavaHelpers.eip3DerivationParent();
         ExtendedSecretKey secretKey = (ExtendedSecretKey) rootSecret.derive(path);
         ExtendedPublicKey pubkey = secretKey.publicKey();
-        P2PKAddress p2pkAddress = JavaHelpers.createP2PKAddress(
-            pubkey.key(),
-            networkType.networkPrefix);
-        return new Address(p2pkAddress);
+
+        return createEip3Address(index, networkType, pubkey);
     }
 
     /**
--- a/common/src/main/java/org/ergoplatform/appkit/Address.java
+++ b/common/src/main/java/org/ergoplatform/appkit/Address.java
@@ -8,6 +8,7 @@ import org.ergoplatform.Pay2SAddress;
 import org.ergoplatform.wallet.secrets.DerivationPath;
 import org.ergoplatform.wallet.secrets.ExtendedPublicKey;
 import org.ergoplatform.wallet.secrets.ExtendedSecretKey;
+
 import scala.util.Try;
 import scorex.util.encode.Base58;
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/ergoplatform/appkit/Bip32Serialization.java
```diff
@@ -0,0 +1,95 @@
...
+
+import scala.collection.immutable.Seq;
+import scorex.crypto.hash.Sha256;
+import scorex.util.encode.Base16$;
+
+/**
+ * Methods to serialize and deserialize key according to bip32 standard
+ * https://en.bitcoin.it/wiki/BIP_0032
+ */
+public class Bip32Serialization {
```
aslesarenko, line 21: Why this is not suitable?
https://github.com/ergoplatform/ergo/blob/93f34e3b9ee6e94505c60d6a5a900ca9d102db86/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedSecretKey.scala#L90

common/src/main/java/org/ergoplatform/appkit/Address.java
```diff
@@ -173,14 +174,35 @@ public static Address createEip3Address(
         // Let's use "m/44'/429'/0'/0/index" path (this path is compliant with EIP-3 which
         // is BIP-44 for Ergo)
         DerivationPath path = JavaHelpers.eip3DerivationWithLastIndex(index);
-        ExtendedSecretKey secretKey = (ExtendedSecretKey)rootSecret.derive(path);
+        ExtendedSecretKey secretKey = (ExtendedSecretKey) rootSecret.derive(path);
```
aslesarenko, line 177: I guess the method below can be reused here to avoid code cloning.

common/src/main/java/org/ergoplatform/appkit/Bip32Serialization.java
```diff
@@ -0,0 +1,95 @@
...
+        if (!Arrays.equals(getPublicHeader(networkType), Arrays.copyOfRange(serializedKey, 0, 4))) {
+            throw new IllegalArgumentException("Given xpubkey did not start with expected version bytes.");
+        }
+
+        return new ExtendedPublicKey(Arrays.copyOfRange(serializedKey, 45, 78),
+            Arrays.copyOfRange(serializedKey, 13, 32 + 13),
+            new DerivationPath((Seq) Constants.eip3DerivationPath().decodedPath().dropRight(1),
+                true));
+    }
+}
```
aslesarenko, line 95: Serializaition roundtrip tests are necessary. See for example https://github.com/ScorexFoundation/sigmastate-interpreter/blob/9e1467ce0a4e981e64246aaeffc35a2efe053978/sigmastate/src/test/scala/org/ergoplatform/ErgoAddressSpecification.scala#L29

aslesarenko (changes requested):

common/src/main/java/org/ergoplatform/appkit/Bip32Serialization.java
MrStahlfelge: It is a viable serialization/deserialization approach, but not compatible to bip-32 and that's how Yoroi, Nautilus and Paper Waller implemented it. So in order to be compatible to the already existing wallets, we need bip-32 compatible serialization and hex encoding.

aslesarenko (approved):

PR #132, opened 2022-02-08
Add additionalRegisters to OutputInfo

MrStahlfelge:
additionalRegisters were omitted in Explorer API but are needed for some information

This was due to a wrong format.

Investigation revealed that the wrong format was also used for DataInput and InputInfo, but it must be changed for all these occasions. 

Examples: 
* [Transaction API with input additionalregisters](https://api.ergoplatform.com/api/v1/transactions/244446acd67a8fc8dbb0e0ee733ffeeebf365e8a62cfcf7329016ab5ac4ffefa)
* [Box API with additionalRegisters](https://api.ergoplatform.com/api/v1/boxes/c58e24560fd04299a3e958cf86b23bdc3eddfeb09cc908bcd80115a049fb25b6)


java-client-generated/src/main/java/org/ergoplatform/explorer/client/model/AdditionalRegister.java
```diff
@@ -0,0 +1,59 @@
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    return serializedValue.equals(((AdditionalRegister) o).serializedValue);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(super.hashCode());
+  }
```
aslesarenko, line 47: The equality implementation is weird.

java-client-generated/src/main/java/org/ergoplatform/explorer/client/model/OutputInfo.java
```diff
@@ -256,23 +251,23 @@ public void setAssets(java.util.List<AssetInstanceInfo> assets) {
...
+  public OutputInfo additionalRegisters(AdditionalRegisters additionalRegisters) {
+    this.additionalRegisters = additionalRegisters;
+    return this;
+  }
+
+   /**
+   * Get additionalRegisters
+   * @return additionalRegisters
+  **/
+  @Schema(required = true, description = "")
```
aslesarenko, line 263: shouldn't it be required = false?

java-client-generated/src/main/java/org/ergoplatform/explorer/client/model/AdditionalRegister.java
MrStahlfelge: To be honest I just copied what the other generated classes had. We can remove it completely when it is me.

java-client-generated/src/main/java/org/ergoplatform/explorer/client/model/OutputInfo.java
MrStahlfelge: Explorer API always returns it, but empty.

aslesarenko (approved):

PR #134, opened 2022-02-09
Eip4 Token support

MrStahlfelge:
[EIP4](https://github.com/ergoplatform/eips/blob/master/eip-0004.md#ergo-tokens-standard) defines a standard how token for NFTs and other asset types should be minted.

This adds an Eip4Token class to appkit which can be used to mint NFTs or fetch information about tokens from Explorer.
So far supported:
Tokens with descriptions, decimals, display name
NFT audio, video, picture, audio with cover image
Membership token is supported as well, but there are no additional getters for its properties and no builder method. Must be added when needed.


common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+
+    @Nullable
+    private byte[] getR7ByteArrayOrNull() {
+        // r7 specifies the asset type and should be Coll<Byte>
+        if (r7 != null && r7.getValue() instanceof Coll) {
+            Coll<?> assetType = (Coll<?>) r7.getValue();
+
+            if (assetType.length() > 0 && assetType.apply(0) instanceof Byte) {
```
aslesarenko, line 119: There is better way of checking the type of Coll elements by calling this method https://github.com/ScorexFoundation/sigmastate-interpreter/blob/d64fd2783c57e70a8ed045a24d2b195aad8eb1f7/library-api/src/main/scala/special/collection/Colls.scala#L374  and then comparing with https://github.com/ScorexFoundation/sigmastate-interpreter/blob/c63e8be72a772c39f1efb828ad1ee89a8deb5a9b/common/src/main/scala/scalan/TypeDesc.scala#L80

Or better yet, wrapping RType into ErgoType and then comparing with ErgoType.byteType()

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
```diff
@@ -0,0 +1,229 @@
...
+        return r8;
+    }
+
+    @Nullable
+    public ErgoValue<?> getMintingBoxR9() {
+        return r9;
+    }
+
+    public enum AssetType {
+        NONE, NFT_PICTURE, NFT_AUDIO, NFT_VIDEO, MEMBERSHIP_THRESHOLD_SIG, UNKNOWN;
```
aslesarenko, line 212: Please add JavaDoc to each enum value and to the whole type and public methods

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
```diff
@@ -0,0 +1,229 @@
...
+        public byte[] getR7ByteArrayForType() {
+            switch (this) {
+                case NFT_PICTURE:
+                    return new byte[]{1, 1};
+                case NFT_AUDIO:
+                    return new byte[]{1, 2};
+                case NFT_VIDEO:
+                    return new byte[]{1, 3};
+                case MEMBERSHIP_THRESHOLD_SIG:
+                    return new byte[]{2, 1};
```
aslesarenko, line 223: Please comment on where this magic values come from.

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
```diff
@@ -0,0 +1,229 @@
...
+        if (r9 != null && isNftAssetType() && r9.getValue() instanceof Tuple2) {
+            return new String(JavaHelpers$.MODULE$.collToByteArray((Coll<Object>) ((Tuple2) r9.getValue())._2),
+                StandardCharsets.UTF_8);
+        } else {
+            return null;
+        }
+    }
+
+    @Nonnull
+    public ErgoValue<Coll<scala.Byte>> getMintingBoxR4() {
```
aslesarenko, line 182: I suggest to rename getNameAsErgoValue, and say in JavaDoc that one use case is to create minting box.
The same for the other methods

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -225,6 +225,31 @@ public UnsignedTransaction putToContractTxUnsigned(
         });
     }
 
+    /**
+     * Creates a new {@link UnsignedTransaction} which sends the given amount of NanoErgs and a
+     * newly minted token to the given contract.
+     *
+     * @param contract     contract to send the newly minted token to
+     * @param tokenBuilder receives the id of the token to mint, must return the new token
+     * @return unsigned transaxtion
```
aslesarenko, line 234:
```suggestion
     * @return unsigned    transaction
```

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -225,6 +225,31 @@ public UnsignedTransaction putToContractTxUnsigned(
...
+     */
+    public UnsignedTransaction mintTokenToContractTxUnsigned(ErgoContract contract, Function<String, Eip4Token> tokenBuilder) {
+        if (!tokensToSpend.isEmpty()) {
+            throw new IllegalArgumentException("Mint token not useable with spending tokens");
```
aslesarenko, line 238:
```suggestion
            throw new IllegalArgumentException("Mint token not possible with spending tokens");
```

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -225,6 +225,31 @@ public UnsignedTransaction putToContractTxUnsigned(
...
+        }
+
+        return buildTxWithDefaultInputs(txB -> {
+            OutBox newBox = txB.outBoxBuilder()
+                .value(amountToSpend)
+                .contract(contract)
+                .mintToken(tokenBuilder.apply(txB.getInputBoxes().get(0).getId().toString()))
```
aslesarenko, line 245: This `txB.getInputBoxes().get(0).getId().toString()` can be encapsulated in a method. It is found in many places

lib-impl/src/main/java/org/ergoplatform/appkit/impl/Eip4TokenBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import retrofit2.Response;
+import special.collection.Coll;
+
+public class Eip4TokenBuilder {
```
aslesarenko, line 26: JavaDoc is missing

lib-impl/src/main/java/org/ergoplatform/appkit/impl/Eip4TokenBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * @param id                  token id
+     * @param amount              token amount
+     * @param additionalRegisters list of registers R4-R9, deserialized to byte[]. Must at least contain R4-R6.
```
aslesarenko, line 30: Describe the encoding which is used to convert bytes to string.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/Eip4TokenBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    @Nonnull
+    public static Eip4Token buildFromRegisters(@Nonnull String id, long amount,
```
aslesarenko, line 33: Suggest to rename buildFromEncodedRegisters

lib-impl/src/main/java/org/ergoplatform/appkit/impl/Eip4TokenBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+            ErgoValue.fromHex(additionalRegisters.get(3)) : null;
+        ErgoValue<?> r8 = (additionalRegisters.size() > 4 && additionalRegisters.get(4) != null) ?
+            ErgoValue.fromHex(additionalRegisters.get(4)) : null;
+        ErgoValue<?> r9 = (additionalRegisters.size() > 5 && additionalRegisters.get(5) != null) ?
+            ErgoValue.fromHex(additionalRegisters.get(5)) : null;
+
+        return new Eip4Token(id, amount, r4, r5, r6, r7, r8, r9);
+    }
+
+    private static ErgoValue<Coll<Object>> getCollByteValue(String serializedValue) {
```
aslesarenko, line 53: suggest renaming to decodeCollByteValue

aslesarenko (changes requested):

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
MrStahlfelge: I have difficulties translating this to Java. Is this code already used from Java methods?

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
MrStahlfelge: Can you give more information? This is the only place I've found that phrase in appkit.

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
MrStahlfelge: The idea was that this name is unaware of the actual content, so registers can be filled without knowing anything about Eip4. If the name was changed, only code aware of the fact that R4 is used to encode the token name can fill the registers.

aslesarenko: But in EIP4 the meaning of the registers is fixed, so why not use the fixed meaning in the name?

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
aslesarenko: search `get(0).getId().toString()` part

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
MrStahlfelge: Two reasons:
* If we do so actual code wanting to fetch the registers (that is for example code for minting the tokens) need to know EIP4 and translate that the name is R4. This is exactly the opposite of encapsulation. Keep in mind that there is also a method for retrieving the token name, which is intended for callers that want to get the name without needing to know how and where it is stored
* Although it is unlikely, the meaning of R4 might be changed or extended in the future. This happened to R9 before, which once was the content link and later became a Pair of content links under certain circumstances. If this happens to R4, all code for minting tokens needs to be changed (which shows that encapsulation is violated)

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
MrStahlfelge: Two matches
* this one
* TxBuilderSpec (which tests if this code worked correctly)

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
aslesarenko: Something like this 
```
    static public <T> ErgoValue<Coll<T>> of(T[] arr, ErgoType<T> tT) {
        Coll<T> value = JavaHelpers.SigmaDsl().Colls().fromArray(arr, tT.getRType());
        if (value.tItem() == ErgoType.byteType().getRType()) {
        ...
        }
        return new ErgoValue<>(value, ErgoType.collType(tT));
    }
```

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
aslesarenko: I don't get it. Eip4Token class is domain specific abstraction - a domain entity which has a set of properties.
Yes, those properties are mapped to registers, because it is how the data is encoded and represented. But this encoding can be hidden from dApps, which use Eip4Token class. (Or it is not the goal? Why?). 
The apps can manipulate with the registers via properties, and which registers are used for that is encapsulated.

aslesarenko (approved):
LGTM, but see the open conversation

common/src/main/java/org/ergoplatform/appkit/Eip4Token.java
MrStahlfelge: Yes, it is the goal to hide internals. To hide the internals from all outside views, we have two methods here:

* `Eip4Token.getTokenName()` is the method to call when a dApp want to retrieve the name of the token. It does not need to know where and how it is stored
* `Eip4Token.getMintingBoxR4()` is the method to call when a dApp wants to mint this token and needs to set register 4 of the minting box, without needing to know what is stored inside the register.

If we rename this method, there is no method available for setting the minting box registers without knowing the internals of what is stored inside the register.

PR #137, opened 2022-02-11
BoxOperations refactoring, add way to ignore boxes spent in mempool for box collection

MrStahlfelge:
BoxOperations get more and more complicated because the methods were static and overloaded to be used with prover, address or list of addresses and list of tokens or no list of tokens. To simplify this, I converted the methods to non-static with object fields holding the properties that can be set and have useful defaults when not set.

getCoveringBoxes method's main logic was moved from BlockChainContextImpl to BoxOperations because it makes more sense to have it there. The method is still available to call from BlockChainContextImpl, but marked Deprecated to make clients call it from BoxOperations.

BoxOperations.createProver methods are still static, but deprecated. These methods don't really fit into the class and are simple helper methods that doesn't need to be provided, so could be removed in the future.

This new structure makes it both simpler to call from clients (see changed tests) and simpler to enhance with more (optional) fields. This is used here to introduce a new field for the source to get the list of unspent boxes from. With this field, we can now override the default implementation that fetches the list of unspent boxes by address from Explorer API. 

This is done with `ExplorerAndPoolUnspentBoxesLoader`, an implementation that omits boxes that are already spent on mempool. This enables us to get a solution for ergoplatform/ergo-wallet-app#5 - moreover, there can be other implementations done for chained transaction support or completely different box selections.

### DIFF
--- a/appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
@@ -2,7 +2,7 @@ package org.ergoplatform.appkit
 
 import java.util
 
-import org.ergoplatform.appkit.BoxOperations.{createProver, loadTop, putToContractTx}
+import org.ergoplatform.appkit.BoxOperations.{createProver}
 import org.ergoplatform.appkit.Parameters.MinFee
 import org.ergoplatform.appkit.impl.ErgoTreeContract
 import org.ergoplatform.appkit.testing.AppkitTesting
@@ -119,7 +119,7 @@ object DhtUtils {
        |  proveDHTuple(groupGenerator, g_y, g_x, g_xy)    // for alice
        |}""".stripMargin);
 
-    val dhtBoxCreationTx = putToContractTx(ctx, sender, false, contract, amountToSend, new util.ArrayList[ErgoToken]())
+    val dhtBoxCreationTx = new BoxOperations(sender, false).withAmountToSpend(amountToSend).putToContractTx(ctx, contract)
     dhtBoxCreationTx
   }
 
@@ -131,7 +131,7 @@ object DhtUtils {
         .registers(ErgoValue.of(g_y), ErgoValue.of(g_xy))
         .build
 
-    val boxesToPayFee = loadTop(ctx, sender.getAddress, MinFee, new util.ArrayList[ErgoToken]())
+    val boxesToPayFee = new BoxOperations(sender.getAddress).loadTop(ctx)
 
     val inputs = new util.ArrayList[InputBox]()
     inputs.add(dhtBox)
--- a/appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
@@ -1,21 +1,21 @@
 package org.ergoplatform.appkit
 
-import java.io.File
-import java.math.BigInteger
-import java.util
-import java.util.Arrays
-
-import org.ergoplatform.{ErgoBox, ErgoScriptPredef}
-import org.ergoplatform.appkit.impl.{ErgoTreeContract, ReducedTransactionImpl, BlockchainContextBase}
+import org.ergoplatform.appkit.InputBoxesSelectionException.NotEnoughErgsException
+import org.ergoplatform.appkit.impl.{BlockchainContextImpl, ErgoTreeContract, ExplorerAndPoolUnspentBoxesLoader}
 import org.ergoplatform.appkit.testing.AppkitTesting
 import org.ergoplatform.restapi.client
-import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
-import org.scalatest.{PropSpec, Matchers}
+import org.ergoplatform.{ErgoBox, ErgoScriptPredef}
 import org.scalacheck.Gen
+import org.scalatest.{Matchers, PropSpec}
+import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
 import scorex.util.ModifierId
-import sigmastate.eval.{CBigInt, CostingBox}
+import sigmastate.eval.CBigInt
 import sigmastate.helpers.NegativeTesting
 import sigmastate.interpreter.HintsBag
+
+import java.io.File
+import java.math.BigInteger
+import java.util.Arrays
 
 class TxBuilderSpec extends PropSpec with Matchers
   with ScalaCheckDrivenPropertyChecks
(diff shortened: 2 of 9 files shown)


MrStahlfelge:
Additional to the tests on this repo, tested by using in ergo-wallet-app and making multiple transactions on the same block https://testnet.ergoplatform.com/en/addresses/3Wwxnaem5ojTfp91qfLw3Y4Sr7ZWVcLPvYSzTsZ4LKGcoxujbxd3

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ExplorerAndPoolUnspentBoxesLoader.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * {@link BoxOperations.IUnspentBoxesLoader} implementation to be used with
+ * {@link BoxOperations#withInputBoxesLoader(BoxOperations.IUnspentBoxesLoader)}
+ * <p>
+ * Instead of the default implementation, this one fetches the mempool from the node connected to
+ * and blacklists the inputs so they aren't used for transactions made by {@link BoxOperations}
+ * methods.
+ * <p>
+ * Note that the mempool is fetched on object instantiation, so don't pass references around as
+ * the data will be outdated after some time.
```
aslesarenko, line 21: Oh, this doesn't sound right. It is better to avoid such effectful operations in a constructor.

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -1,119 +1,159 @@
...
 import java.util.List;
+import java.util.function.Function;
 
-import static com.google.common.base.Preconditions.checkState;
-import static org.ergoplatform.appkit.Parameters.MinFee;
+import javax.annotation.Nonnull;
 
 /**
- * A collection of utility operations implemented in terms of abstract Appkit interfaces.
+ * A collection of utility operations for selecting boxes and building transactions
```
aslesarenko, line 20: This was the key point actually. By restricting the implementation to be on top of interfaces only, decouples this methods' logic from the implementation of the core Appkit interfaces.
A property worth preserving.

aslesarenko (changes requested):

appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
```diff
@@ -131,7 +131,7 @@ object DhtUtils {
         .registers(ErgoValue.of(g_y), ErgoValue.of(g_xy))
         .build
 
-    val boxesToPayFee = loadTop(ctx, sender.getAddress, MinFee, new util.ArrayList[ErgoToken]())
+    val boxesToPayFee = new BoxOperations(sender.getAddress).loadTop(ctx)
```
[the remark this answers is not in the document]
MrStahlfelge: We can do that, but it does not really make any change in my opinion: Will reduce constructors to a single one, but the create method is overloaded.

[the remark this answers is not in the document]
aslesarenko: Yeah, but it is more idiomatic for builders implementation.

appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
```diff
@@ -317,6 +314,41 @@ class TxBuilderSpec extends PropSpec with Matchers
...
+        storage.unlock("abc")
+
+        val recipient = address
+
+        val amountToSend = 1000000
+        val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script)
+
+        val senders = Arrays.asList(storage.getAddressFor(NetworkType.MAINNET))
+        val unsigned = new BoxOperations(senders).withAmountToSpend(amountToSend)
+          .withInputBoxesLoader(new ExplorerAndPoolUnspentBoxesLoader(ctx.asInstanceOf[BlockchainContextImpl]))
```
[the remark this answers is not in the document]
aslesarenko: One problem with ErgoNodeFacade is that it exposes (or I would say leaks) Retrofit in the signatures. Ideally all interactions with API should be done via abstracted interfaces, where Retrofit (as an implementation detail) is not exposed.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ExplorerAndPoolUnspentBoxesLoader.java
aslesarenko: I think we are dealing with Provider/Loader pattern here. It is better to keep the Loader itself stateless, so it only encapsulate the loading logic itself, but doesn't preserve the loaded data (which is another responsibility). The you can share the loader as necessary.

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
aslesarenko: The idea is to clearly decouple Impl classes (or hide them) from the code, which uses abstract interfaces. Basically any code that uses interfaces alone is abstracted from the concrete detail of how those interfaces are implemented in Impl classes. 
Current implementation of BoxOperations is good enough, as it doesn't use Impl classes and is written on top of the interfaces.
But I suggest to leave the phrase `implemented in terms of abstract Appkit interfaces.` in place.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ExplorerAndPoolUnspentBoxesLoader.java
MrStahlfelge: That would mean to fetch the data from node on any page, or do an `if (page == 0) { fetch stuff }` workaround. I can change it like that but really think this is hacky.
I had planned to implement transaction chaining as well, and that situation is not even to handle that way. Unconfirmed boxes need to be fetched on the last page, so when explorer returns 0 boxes. To not run into a never ending recursion, the loader must save that it already returned the unconfirmed boxes to not return it again. If you insist on a stateless loader in appkit, I can't implement the feature here but have to do it on wallet-app repo. Okay for me, but won't help others to use the feature.

appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
```diff
@@ -254,9 +254,8 @@ class TxBuilderSpec extends PropSpec with Matchers
 
       val recipient = senderProver.getEip3Addresses.get(1)
       val amountToSend = 1000000
-      val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script)
-      val signed = BoxOperations.putToContractTx(ctx,
-          senderProver, false, pkContract, amountToSend, new util.ArrayList[ErgoToken]())
+      val signed = new BoxOperations(senderProver, false).withAmountToSpend(amountToSend)
+        .send(ctx, recipient)
```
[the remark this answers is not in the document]
MrStahlfelge: On second check, I found that this really sends the transaction - so the name is correct.

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
MrStahlfelge: Done.

appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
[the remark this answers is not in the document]
MrStahlfelge: Done.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ExplorerAndPoolUnspentBoxesLoader.java
MrStahlfelge: I have now changed the interface, added `prepare()` and `prepareForAddress` so any preparation needed can be done right in time. This is more flexible, but the implementation for pool support still needs to hold a state between the calls to `prepare()` and `loadBoxesPage()`.

appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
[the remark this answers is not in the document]
MrStahlfelge: After thinking some more about it, I would leave it as it is. 
* I agree that exposing the Retrofit stuff on the interface is not a good idea. 
* On the other hand, adding the needed methods on the interface makes it more cluttered and after all, other implementations of the interface will only throw an `UnsupportedOperationException` for these methods
* I think it is okay to have the dependency on BlockchainContextImpl as the `ExplorerAndPoolUnspentBoxesLoader` is an optional component sitting in the impl package. It does not make sense to use it for other types of contexts, or another implementation should be provided.

aslesarenko (approved):

MrStahlfelge:
For the time being, we leave it as it is. When #118 is resolved, we will add an abstract api client layer that can be implemented by the implementation layer. ExplorerAndPoolUnspentBoxesLoader will then be able to use this abstract layer.

PR #138, opened 2022-02-15
Making  public get unsigne dtx for UnsignedTransactionImpl

dav009:
👋  I am updating the playground with a SimulatedBlockchainContext. 

having this function as public would let me delete a lot of code.


MrStahlfelge:
On this occasion, I would like to ask if this method and the other getters could be added to the `UnsignedTransaction` interface. I always wondered why it is empty and need to access some of the information in my clients (for me it is `getBoxesToSpend()`.

aslesarenko (approved):

PR #140, opened 2022-02-15
New Accessor Methods

K-Singh:
Created additional accessor methods in InputBox, OutBox, ErgoContract, and UnsignedTransaction along with their Impl versions. Also replaced token function in OutBox with getTokens() and passed in additional variables when building UnsignedTransactions

### DIFF
--- a/appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
@@ -270,7 +270,7 @@ class TxBuilderSpec extends PropSpec with Matchers
       val recipient = address
 
       val amountToSend = 1000000
-      val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script)
+      val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script, recipient.getNetworkType)
 
       val senders = Arrays.asList(storage.getAddressFor(NetworkType.MAINNET))
       val unsigned = BoxOperations.createForSenders(senders).withAmountToSpend(amountToSend)
@@ -337,7 +337,7 @@ class TxBuilderSpec extends PropSpec with Matchers
         val recipient = address
 
         val amountToSend = 1000000
-        val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script)
+        val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script, recipient.getNetworkType)
 
         val senders = Arrays.asList(storage.getAddressFor(NetworkType.MAINNET))
         val unsigned = BoxOperations.createForSenders(senders).withAmountToSpend(amountToSend)
--- a/common/src/main/java/org/ergoplatform/appkit/impl/ErgoTreeContract.java
+++ b/common/src/main/java/org/ergoplatform/appkit/impl/ErgoTreeContract.java
@@ -1,7 +1,10 @@
 package org.ergoplatform.appkit.impl;
 
+import org.ergoplatform.ErgoAddress;
+import org.ergoplatform.appkit.Address;
 import org.ergoplatform.appkit.Constants;
 import org.ergoplatform.appkit.ErgoContract;
+import org.ergoplatform.appkit.NetworkType;
 import sigmastate.Values;
 
 public class ErgoTreeContract implements ErgoContract {
@@ -30,4 +33,8 @@ public class ErgoTreeContract implements ErgoContract {
     public Values.ErgoTree getErgoTree() {
         return _ergoTree;
     }
+
+    @Override
+    public Address getAddress(NetworkType networkType) { return Address.fromErgoTree(_ergoTree, networkType); }
+
 }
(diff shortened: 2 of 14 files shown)


lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxImpl.scala
```diff
@@ -1,16 +1,27 @@
...
   override def getValue: Long = _ergoBoxCandidate.value
 
   override def getCreationHeight: Int = _ergoBoxCandidate.creationHeight
 
-  override def token(id: ErgoId): ErgoToken = null
+  override def getTokens: util.List[ErgoToken] = Iso.isoTokensListToPairsColl.from(_ergoBoxCandidate.additionalTokens)
+
+  override def getRegisters: util.List[ErgoValue[_]] = {
+    val registers = _ergoBoxCandidate.additionalRegisters.values.toIndexedSeq
+    Iso.JListToIndexedSeq(Iso.isoErgoValueToSValue).from(registers)
```
aslesarenko, line 19: Just a hint, is Scala you can use convertTo method when you know the resulting type. In most of the cases the compiler will summon the necessary iso automatically.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxImpl.scala
```diff
@@ -1,16 +1,27 @@
...
+  }
+
+  override def getBytes: Array[Byte] = _ergoBoxCandidate.bytesWithNoRef
```
aslesarenko, line 22: I suggest to rename the method `getBytesWithoutRef` to be consistent with Sigma (I don't like the original `bytesWithNoRef`, but it is there historically and here we can change it a bit).

And also add `getBytes` method which will return the full bytes.

aslesarenko (changes requested):
in general, it looks good and in the spirit of Appkit design.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxImpl.scala
K-Singh: Yes, was in java mode while writing some of these methods 😄 
I'll try to make this look a bit cleaner.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxImpl.scala
K-Singh: I am not sure how to get the full bytes for ErgoBoxCandidate, the only method in sigmastate that I can see is bytesWithNoRef. I am not sure that it is possible to get the full bytes if there is no transaction reference data, so should I be inputting placeholder values to get the full bytes?

common/src/main/java/org/ergoplatform/appkit/impl/ErgoTreeContract.java
```diff
@@ -30,4 +33,8 @@ public ErgoTreeContract substConstant(String name, Object value) {
     public Values.ErgoTree getErgoTree() {
         return _ergoTree;
     }
+
+    @Override
+    public Address getAddress(NetworkType networkType) { return Address.fromErgoTree(_ergoTree, networkType); }
```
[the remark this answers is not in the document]
aslesarenko: Yeah, networkType is used by the compiler to decode PK("address") expressions. See https://github.com/ScorexFoundation/sigmastate-interpreter/blob/a3068e8b7c497a6dd1b7871708a7473c5fe0c830/sigmastate/src/main/scala/sigmastate/lang/SigmaPredef.scala#L153
And networkType is not part of ErgoTree, so this asymmetry between ErgoContract implementations because of the compiler.
I would suggest to remove this methods, as they don't encapsulate much, but create confusion and ambiguity.
But as you mentioned, it would be useful to have direct method `ErgoContract.getAddress` because yes,  contracts and addresses are kind of the same thing.
However, this method will require passing networkPrefix to the ErgoTreeContract constructor, and I think that is ok and it can be added.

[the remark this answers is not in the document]
K-Singh: Yeah, passing in NetworkType into ErgoTreeContract seems like the best way to go about this. I'll work on implementing this today.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxImpl.scala
aslesarenko: Let's leave getBytes for another PR.

K-Singh:
Added the the changes you requested @aslesarenko

appkit/src/test/scala/org/ergoplatform/appkit/TxBuilderSpec.scala
```diff
@@ -270,7 +270,7 @@ class TxBuilderSpec extends PropSpec with Matchers
       val recipient = address
 
       val amountToSend = 1000000
-      val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script)
+      val pkContract = new ErgoTreeContract(recipient.getErgoAddress.script, recipient.getNetworkType)
```
aslesarenko, line 273: Looks like now a new constructor can be added with Address parameter.

aslesarenko (approved):
LGTM, but please the new comment I added

PR #141, opened 2022-02-22
BoxOperations: Consider amount for change box

MrStahlfelge:
When we have tokens in selected inboxes that shouldn't be sent, a change box is needed to return the tokens to the sending address. That means that a small amount of ERG is needed on top of the actual amount to send which must be considered when selecting boxes.

This change adds the ability to BoxOperations to account for a change box amount and an accompanying test case. However, the test case files although BoxOperations is now correctly selecting the two boxes of the test case. This is due to `DefaultBoxSelector` ignoring the second, needed box which needs to be fixed in ergo-wallet.

Follow-up to #103

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/SelectTokensHelper.java
+++ b/common/src/main/java/org/ergoplatform/appkit/SelectTokensHelper.java
@@ -6,7 +6,8 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Helper class to keep track of amount of tokens to spend and amount of tokens already found
+ * Helper class to keep track of amount of tokens to spend and tokens already covered by boxes.
+ * Used to determine if more and which boxes need to be selected, and if a change box is needed.
  */
 public class SelectTokensHelper {
     private final HashMap<String, Long> tokensLeft;
@@ -44,16 +45,17 @@ public class SelectTokensHelper {
     }
 
     /**
-     * Marks the given tokens as needed, subtracting the amount from the remaining amount of
-     * tokens to find.
-     * Also updates if a change box is needed
+     * Marks the given tokens as selected, subtracting the amount values from the remaining amount
+     * of tokens needed to fulfill the initial tokens to spend.
+     * Also keeps track if a change box is needed in case we selected to many tokens, see
+     * {@link #isChangeBoxNeeded()}
      */
-    public void useTokens(Iterable<ErgoToken> foundTokens) {
-        for (ErgoToken foundToken : foundTokens) {
-            String tokenId = foundToken.getId().toString();
+    public void useTokens(Iterable<ErgoToken> selectedTokens) {
+        for (ErgoToken selectedToken : selectedTokens) {
+            String tokenId = selectedToken.getId().toString();
             if (tokensLeft.containsKey(tokenId)) {
                 Long currentValue = tokensLeft.get(tokenId);
-                long newValue = currentValue - foundToken.getValue();
+                long newValue = currentValue - selectedToken.getValue();
                 tokensLeft.put(tokenId, newValue);
                 if (newValue < 0) {
                     changeBoxNeeded = true;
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/ergoplatform/appkit/SelectTokensHelper.java
```diff
@@ -10,33 +10,60 @@
...
         }
         return tokensNeeded;
     }
 
+    /**
+     * Marks the given tokens as needed, subtracting the amount from the remaining amount of
+     * tokens to find.
+     * Also updates if a change box is needed
+     */
+    public void useTokens(Iterable<ErgoToken> foundTokens) {
```
aslesarenko, line 51: Please update the doc. It is not clear how tokens are marked. What is means "tokens to find"? Find where?

common/src/main/java/org/ergoplatform/appkit/SelectTokensHelper.java
```diff
@@ -58,4 +85,8 @@ public boolean areTokensCovered() {
         }
         return result;
     }
+
+    public boolean isChangeBoxNeeded() {
```
aslesarenko, line 89: Please add JavaDoc.

common/src/main/java/org/ergoplatform/appkit/SelectTokensHelper.java
```diff
@@ -10,33 +10,60 @@
...
+                if (newValue < 0) {
+                    changeBoxNeeded = true;
+                }
+            } else {
+                changeBoxNeeded = true;
+            }
+        }
+    }
+
     public boolean areTokensCovered() {
```
aslesarenko, line 67: Please add JavaDoc

common/src/main/java/org/ergoplatform/appkit/SelectTokensHelper.java
```diff
@@ -10,33 +10,60 @@
  */
 public class SelectTokensHelper {
```
aslesarenko, line 11: Please clarify the description. What "tokens already found" means?

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -157,18 +159,27 @@ public String send(
...
 
         for (Address sender : senders) {
             inputBoxesLoader.prepareForAddress(sender);
-            CoveringBoxes unspent = getCoveringBoxesFor(remainingAmount, remainingTokens,
+            CoveringBoxes addressUnspentBoxes = getCoveringBoxesFor(remainingAmount, remainingTokens,
+                changeBoxConsidered,
                 page -> inputBoxesLoader.loadBoxesPage(ctx, sender, page));
-            for (InputBox b : unspent.getBoxes()) {
+
+            // change box needed but not considered yet?
```
aslesarenko, line 174: What it means "considered"? Maybe different verb will serve better.

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -255,13 +266,15 @@ public static SignedTransaction spendBoxesTx(
      * - returning an empty list means the source of input boxes is drained and no further page will
      *   be loaded
      *
-     * @param amountToSpend    amount of NanoErgs to be covered
-     * @param tokensToSpend    ErgoToken to spent
-     * @param inputBoxesLoader method returning paged sets of InputBoxes, see above
+     * @param amountToSpend       amount of NanoErgs to be covered
+     * @param tokensToSpend       ErgoToken to spent
+     * @param changeBoxConsidered true if change box is already considered by amountToSpend
```
aslesarenko, line 271: the word `considered` is not clear.

aslesarenko (changes requested):

MrStahlfelge:
Addressed the feedback, please check

aslesarenko (approved):

PR #144, opened 2022-02-24
Set of small additions

MrStahlfelge:
This is a set of small additions in preparation of Eip4Token branch.

* `ErgoValue.pairOf` helper was missing
* `collToByteArray`, a Java-friendly way to convert Coll[Byte] to byte[]
* `IErgoBox` super interface for InputBox and Outbox

I am aware that `IErgoBox` does not follow the normal naming style, but `ErgoBox` was not possible as it clashed with ErgoBox.scala - better suggestions welcome

### DIFF
--- a/lib-api/src/main/java/org/ergoplatform/appkit/IErgoBox.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/IErgoBox.java
@@ -1,0 +1,34 @@
+package org.ergoplatform.appkit;
+
+import java.util.List;
+
+import sigmastate.Values;
+
+public interface IErgoBox {
+    /**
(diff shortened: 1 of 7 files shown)


lib-api/src/main/java/org/ergoplatform/appkit/IErgoBox.java
```diff
@@ -0,0 +1,34 @@
+package org.ergoplatform.appkit;
+
+import java.util.List;
+
+import sigmastate.Values;
+
+public interface IErgoBox {
```
aslesarenko, line 7: Better TxBox or TransactionBox.
Please add JavaDoc including the motivation for adding interface.

aslesarenko (approved):

PR #145, opened 2022-03-06
Add Address.getErgoContract convenience method

MrStahlfelge:
This one is called a lot within the project and in client projects, so a helper method is justified.

### DIFF
--- a/appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
@@ -127,7 +127,7 @@ object DhtUtils {
     val tx = BoxOperations.createForSender(sender.getAddress, ctx).buildTxWithDefaultInputs { txB: UnsignedTransactionBuilder =>
       val outBox = txB.outBoxBuilder
         .value(dhtBox.getValue)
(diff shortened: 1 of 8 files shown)


appkit/src/test/scala/org/ergoplatform/appkit/AnonymousAccessSpec.scala
```diff
@@ -127,7 +127,7 @@ object DhtUtils {
     val tx = BoxOperations.createForSender(sender.getAddress, ctx).buildTxWithDefaultInputs { txB: UnsignedTransactionBuilder =>
       val outBox = txB.outBoxBuilder
         .value(dhtBox.getValue)
-        .contract(new ErgoTreeContract(receiver.getErgoAddress.script, receiver.getNetworkType))
+        .contract(receiver.getErgoContract)
```
aslesarenko, line 130: Better name would be `asErgoContract` or `toErgoContract`, because Address and ErgoContract are isomorphic. Basically two different representations of the same thing.

aslesarenko (commented):
Consider adding also the reverse method ErgoContract.toAddress

MrStahlfelge:
Renamed both

aslesarenko (commented):
Please also add test vectors for new toContract and toAddress methods.
The methods should be mutually inverse.
So you can use Scala property testing and ErgoTree generators to test roundtrip property.

MrStahlfelge:
Done

PR #146, opened 2022-03-06
Save fat jar for releases

MrStahlfelge:
This stores the fat jar for releases, so people don't need to build it themselves.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ env:
 
 jobs:
   publish_release:
-    name: Publish release to Sonatype
+    name: Publish release
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,3 +42,11 @@ jobs:
...
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+
+      - name: Build fat jar
+        run: sbt clean assembly
+      - name: Save fat jar artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ergo-appkit.jar
+          path: target/**/ergo-appkit-*.jar
```
aslesarenko, line 52: What this action do?

MrStahlfelge: Takes every ergo-appkit-*.jar file (the fat jar produced by sbt assembly) and saves it as an artifact on the release action, so everyone can download and use it without building on their own.
Convenient for Python users not wanting to set up sbt.

PR #148, opened 2022-03-09
Make verifySignature static on ErgoProver

MrStahlfelge:
verifying a signature is independant from actual prover and doesn't need any secrets, so this changes the code towards being usable without constructing a prover holding secrets.

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/Signature.java
+++ b/common/src/main/java/org/ergoplatform/appkit/Signature.java
@@ -14,7 +14,7 @@ public class Signature {
     /**
      * Verifies a signature on given (arbitrary) message for a given public key.
      *
-     * @param sigmaTree     public key (represented as a tree)
+     * @param sigmaTree     public key (represented as a sigma proposition tree)
      * @param message       message to verify
      * @param signature signature for the message
      * @return whether signature is valid or not
@@ -24,7 +24,7 @@ public class Signature {
     }
 
     /**
-     * Verifies a signature on given (arbitrary) message for a
+     * Verifies a signature on given (arbitrary) message
      * using an address' public key.
      *
      * @param addr          address whose public key will be used to verify message
--- a/common/src/main/java/org/ergoplatform/appkit/Signature.java
+++ b/common/src/main/java/org/ergoplatform/appkit/Signature.java
@@ -1,0 +1,38 @@
+package org.ergoplatform.appkit;
+
+import org.ergoplatform.P2PKAddress;
+import org.ergoplatform.ErgoLikeInterpreter;
+
+import sigmastate.Values.SigmaBoolean;
+import sigmastate.eval.CompiletimeIRContext;
+
+public class Signature {
+    private Signature() {
+        // prevent instantiation
+    }
+
+    /**
+     * Verifies a signature on given (arbitrary) message for a given public key.
+     *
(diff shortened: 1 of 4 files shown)


lib-api/src/main/java/org/ergoplatform/appkit/ErgoProver.java
```diff
@@ -94,22 +97,26 @@
...
      *
-     * @param sigmaTree public key (represented as a tree)
-     * @param message message to verify
+     * @param sigmaTree     public key (represented as a tree)
+     * @param message       message to verify
      * @param signedMessage signature for the message
      * @return whether signature is valid or not
      */
-    boolean verifySignature(SigmaBoolean sigmaTree, byte[] message, byte[] signedMessage);
+    static boolean verifySignature(SigmaBoolean sigmaTree, byte[] message, byte[] signedMessage) {
```
aslesarenko, line 105: signedMessage name is misleading, better call it `signature` as in the original method of interpreter.

aslesarenko (changes requested):

lib-api/src/main/java/org/ergoplatform/appkit/ErgoProver.java
```diff
@@ -94,22 +97,26 @@
...
-     * @param message message to verify
+     * @param addr          address whose public key will be used to verify message
+     * @param message       message to verify
      * @param signedMessage signature for the message
      * @return whether signature is valid or not
      */
-    boolean verifySignature(P2PKAddress addr, byte[] message, byte[] signedMessage);
+    static boolean verifySignature(P2PKAddress addr, byte[] message, byte[] signedMessage) {
+        return new ErgoLikeInterpreter((new CompiletimeIRContext())).verifySignature(addr.pubkey(), message, signedMessage);
+    }
```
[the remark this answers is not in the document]
MrStahlfelge: Done, as discussed on Discord AppkitProvingInterpreter needs secrets that are not used here.

common/src/main/java/org/ergoplatform/appkit/Signature.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class Signature {
+    private Signature() {
+        // prevent instantiation
+    }
+
+    /**
+     * Verifies a signature on given (arbitrary) message for a given public key.
+     *
+     * @param sigmaTree     public key (represented as a tree)
```
aslesarenko, line 17:
```suggestion
     * @param sigmaTree     public key (represented as a sigma proposition tree)
```

common/src/main/java/org/ergoplatform/appkit/Signature.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param signature signature for the message
+     * @return whether signature is valid or not
+     */
+    public static boolean verifySignature(SigmaBoolean sigmaTree, byte[] message, byte[] signature) {
+        return new ErgoLikeInterpreter((new CompiletimeIRContext())).verifySignature(sigmaTree, message, signature);
+    }
+
+    /**
+     * Verifies a signature on given (arbitrary) message for a
+     * using an address' public key.
```
aslesarenko, line 28:
```suggestion
     * Verifies a signature on given (arbitrary) message
     * using an address' public key.
```

aslesarenko (approved):
LGTM, please also apply edit suggestions

PR #156, opened 2022-03-31
Cleanup ExplorerApiClient and ApiClient, share OkHttp instance

MrStahlfelge:
This cleans up ExplorerApiClient from a lot of unreachable code, and some code smells in ApiClient when using a proxy.

NodeAndExplorerDataSourceImpl shares its OkHttpClient between both ApiClient and ExplorerApiClient, as it is strongly advised to use OkHttpClient as a singleton (see OkHttpClient documentation). So instead of two new instances we only use one this way.

To get this even better, it should be possible to set the OkHttpClient to be used by `RestApiErgoClient` so that not a new one has to be spin up for every instantiation of it. I think the current proxy param of RestApiErgoClient should be replaced by a OkHttpClientBuilder param as this lets clients set proxy and many more settings like timeout etc. I just wanted to ask you on your opinion @aslesarenko because this means leaking implementation details to the clients, but this has strong technical reasons and is done by Retrofit for the same reason.

### DIFF
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
@@ -59,9 +59,9 @@ public class NodeAndExplorerDataSourceImpl implements BlockchainDataSource {
 
     public NodeAndExplorerDataSourceImpl(ApiClient nodeClient, @Nullable ExplorerApiClient explorerClient) {
 
-        OkHttpClient _ok = nodeClient.getOkBuilder().build();
+        OkHttpClient ok = nodeClient.getOkBuilder().build();
         Retrofit nodeRetrofit = nodeClient.getAdapterBuilder()
-            .client(_ok)
+            .client(ok)
             .build();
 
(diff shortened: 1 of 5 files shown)


lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
```diff
@@ -59,6 +59,19 @@
 
     public NodeAndExplorerDataSourceImpl(ApiClient nodeClient, @Nullable ExplorerApiClient explorerClient) {
 
+        if (explorerClient != null) {
+            OkHttpClient okExplorer = explorerClient.getOkBuilder().build();
+            Retrofit _retrofitExplorer = explorerClient.getAdapterBuilder()
```
aslesarenko, line 64: Usage of `_` prefix is reserved for member fields.

aslesarenko:
@MrStahlfelge I completely agree with making singleton and suggested refactorings as long as interfaces are free from implementation details.

MrStahlfelge:
Great that made it easier!

Ping to @anon2020s since he added the proxy param

aslesarenko (approved):

PR #157, opened 2022-04-05
ErgoAuthUtils added for EIP-28

MrStahlfelge:
Adds ErgoAuthUtils utility methods for ErgoAuth ergoplatform/eips#53

On this occasion, ColdErgoClient was improved so that it does not need to be instantiated by using Explorer API model classes.

### DIFF
--- a/appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
@@ -1,0 +1,41 @@
+package org.ergoplatform.appkit
+
+import org.scalatest.{Matchers, PropSpec}
+import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
+import scorex.util.Random
+import sigmastate.interpreter.HintsBag
+
+import java.nio.charset.StandardCharsets
+
+class ErgoAuthSpec extends PropSpec with Matchers with ScalaCheckDrivenPropertyChecks
+  with AppkitTestingCommon {
+
+  property("ErgoAuth address roundtrip") {
+    // server side
+    val serializedSigmaBoolean = ErgoAuthUtils.serializeSigmaBoolean(address)
+    // message to sign should be something random and not repeating
+    val requestedMessage = addrStr + System.currentTimeMillis().toString
+
+    // transferred to client, and now we are on client side
+    val signedMessage = new String(Random.randomBytes(16)) + requestedMessage +
+      new String(Random.randomBytes(32))
+    val signature = new ColdErgoClient(address.getNetworkType, Parameters.ColdClientMaxBlockCost)
+      .execute { ctx =>
+
+        val prover = ctx.newProverBuilder().withMnemonic(mnemonic, SecretString.empty()).build()
+        prover.signMessage(ErgoAuthUtils.deserializeSigmaBoolean(serializedSigmaBoolean),
+          signedMessage.getBytes(StandardCharsets.UTF_8),
+          HintsBag.empty)
+      }
+
+    // transferred to server...
+    ErgoAuthUtils.verifyResponse(address, requestedMessage, signedMessage, signature) shouldBe (true)
+
+    // and in case someone wanted to fool us
+    ErgoAuthUtils.verifyResponse(address,
+      requestedMessage,
+      signedMessage,
+      new Array[Byte](0)) shouldBe (false)
+  }
+
+}
--- a/appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
@@ -11,31 +11,35 @@ class ErgoAuthSpec extends PropSpec with Matchers with ScalaCheckDrivenPropertyC
   with AppkitTestingCommon {
 
   property("ErgoAuth address roundtrip") {
-    // server side
-    val serializedSigmaBoolean = ErgoAuthUtils.serializeSigmaBoolean(address)
+    val sigmaPropFromAddress = SigmaProp.createFromAddress(address)
+    // ---- server side ----
+    val serializedSigmaBoolean = sigmaPropFromAddress.toBytes
     // message to sign should be something random and not repeating
     val requestedMessage = addrStr + System.currentTimeMillis().toString
 
-    // transferred to client, and now we are on client side
+    // ---- transferred to client, and now we are on client side ----
+
+    // EIP-28: "the wallet app adds some own bytes to the obtained message from ErgoAuthRequest"
     val signedMessage = new String(Random.randomBytes(16)) + requestedMessage +
       new String(Random.randomBytes(32))
     val signature = new ColdErgoClient(address.getNetworkType, Parameters.ColdClientMaxBlockCost)
       .execute { ctx: BlockchainContext =>
 
         val prover = ctx.newProverBuilder().withMnemonic(mnemonic, SecretString.empty()).build()
-        prover.signMessage(ErgoAuthUtils.deserializeSigmaBoolean(serializedSigmaBoolean),
+        prover.signMessage(SigmaProp.parseFromBytes(serializedSigmaBoolean),
           signedMessage.getBytes(StandardCharsets.UTF_8),
           HintsBag.empty)
       }
 
-    // transferred to server...
-    ErgoAuthUtils.verifyResponse(address, requestedMessage, signedMessage, signature) shouldBe (true)
+    // ---- transferred to server... ----
+    ErgoAuthUtils.verifyResponse(sigmaPropFromAddress,
+      requestedMessage, signedMessage, signature) shouldBe true
 
     // and in case someone wanted to fool us
-    ErgoAuthUtils.verifyResponse(address,
+    ErgoAuthUtils.verifyResponse(sigmaPropFromAddress,
       requestedMessage,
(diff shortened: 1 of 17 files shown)


common/src/main/java/org/ergoplatform/appkit/ErgoAuthUtils.java
```diff
@@ -0,0 +1,62 @@
...
+     * Serializes the given SigmaBoolean to be sent in ErgoAuthRequest.
+     */
+    public static byte[] serializeSigmaBoolean(SigmaBoolean sigmaBoolean) {
+        return Iso.isoSigmaBooleanToByteArray().to(sigmaBoolean);
+    }
+
+    /**
+     * Converts given address to SigmaBoolean and serializes it  to be sent in ErgoAuthRequest.
+     */
+    public static byte[] serializeSigmaBoolean(Address address) {
```
aslesarenko, line 22: This method creates confusion. It looks like overload of the above method, implying that Address and SigmaBoolean are the same (or isomorphic), but they are far from that.
It is better to remove this method and instead add Address.getSigmaBoolean method. 
Also Address.isSigmaBoolean is necessary, because not every Address is backed by SigmaBoolean. 
Note p2pk is a specific case of SigmaBoolean address.

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  property("ErgoAuth address roundtrip") {
+    // server side
+    val serializedSigmaBoolean = ErgoAuthUtils.serializeSigmaBoolean(address)
+    // message to sign should be something random and not repeating
+    val requestedMessage = addrStr + System.currentTimeMillis().toString
+
+    // transferred to client, and now we are on client side
+    val signedMessage = new String(Random.randomBytes(16)) + requestedMessage +
+      new String(Random.randomBytes(32))
```
aslesarenko, line 21: What these random strings represent?

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+      .execute { ctx: BlockchainContext =>
+
+        val prover = ctx.newProverBuilder().withMnemonic(mnemonic, SecretString.empty()).build()
+        prover.signMessage(ErgoAuthUtils.deserializeSigmaBoolean(serializedSigmaBoolean),
+          signedMessage.getBytes(StandardCharsets.UTF_8),
+          HintsBag.empty)
+      }
+
+    // transferred to server...
+    ErgoAuthUtils.verifyResponse(address, requestedMessage, signedMessage, signature) shouldBe (true)
```
aslesarenko, line 32: Again, it is better to use SigmaBoolean here instead of Address. SigmaBoolean is very important concept which sits between p2pk and arbitrary Address. It is more general than public key, but less general then Address. 
It is better to make this distinction explicit everywhere in Appkit.
I would even suggest to add SigmaProp class next to Address class and wrap SigmaBoolean into it.
NOTE, there is special.sigma.SigmaProp interface in Sigma interpreter which represent exactly such wrapping (see implementation in sigmastate.eval.CSigmaProp)

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+import sigmastate.interpreter.HintsBag
+
+import java.nio.charset.StandardCharsets
+
+class ErgoAuthSpec extends PropSpec with Matchers with ScalaCheckDrivenPropertyChecks
+  with AppkitTestingCommon {
+
+  property("ErgoAuth address roundtrip") {
+    // server side
+    val serializedSigmaBoolean = ErgoAuthUtils.serializeSigmaBoolean(address)
```
aslesarenko, line 15: It is better to use SigmaBoolean here and also below.

aslesarenko (changes requested):
I suggest to introduce SigmaProp wrapper (see one of the comments), move related serialization code there and use it where necessary instead of Address.

MrStahlfelge:
I have made the requested additions, but please note: 

* the new Address methods have a very low level of originality, they are only duplications for `isP2PK()`/`getPublicKey()` 
* `ErgoAuthUtils` does only hold a very basic check nonw and could be removed completely. The question the dApp developers reviewing the EIP all asked was "What is a SigmaBoolean, we want to check if a user can access a certain address holding a NFT.". The utility class does not help with this problem any more.

common/src/main/java/org/ergoplatform/appkit/Address.java
```diff
@@ -129,6 +129,21 @@ public ErgoContract toErgoContract() {
...
+        return isP2PK();
+    }
+
+    /**
+     * @return SigmaBoolean value of this address. Throws an error if
+     * {@link #isSigmaBoolean()} is false
+     */
+    public Values.SigmaBoolean getSigmaBoolean() {
+        return getPublicKey();
+    }
```
aslesarenko, line 145: It is ok for a stub. The actual implementation should check that the ErgoTree corresponding to this address is equivalent to this special case https://github.com/ScorexFoundation/sigmastate-interpreter/blob/23d4585adc999101bd393fc9c9747b1e858b0e35/sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala#L254

The logic for extracting the `prop` in the code fragment above which can be used in Appkit is simple
```scala
val prop = ergoTree.toProposition(ergoTree.isConstantSegregation)
```
Then you can check the prop is SigmaPropConstant and extract SigmaBoolean from it the same way it is done in the interpreter.

As this is very fundamental part of ErgoTree semantics I think you need to know it.

aslesarenko (changes requested):

aslesarenko:
> * we want to check if a user can access a certain address holding a NFT

In blockchain ownership of a box is proved by showing that "the owner" can spend it.
This can be done in general, without knowledge of specific contract and without the actual spending. Will describe it below. But this is not how ErgoAuthUtils were implemented.
 
The User and the dApp follow this steps in order to authenticate access to The Box:
1) User creates a special AuthMessage which is like ErgoTransaction, but not quite, it:
   - contains the Box as a single input
   - doesn't contain outputs
   - is serializable to bytes which will be signed
 2) The user executes the Box.ergoTree in the current blockchain context, reduces it to SigmaProp and then signs the AuthMessage.bytes with it. 
(this step is similar to signing a transaction in Appkit)
3) The user sends SignedAuthMessage to the dApp
4) The dApp:
  - receives SignedAuthMessage
  - extracts the Box from it
  - reduces Box.ergoTree to SigmaProp
  - extracts SignedAuthMessage.signature
  - verifies signature against SigmaProp and the AuthMessage.bytes
 (this step is similar to what each Ergo node does in verifying a transaction)

If we want to implement this protocol, we can basically clone Transaction and SignedTransaction implementation and simplify it.

common/src/main/java/org/ergoplatform/appkit/Address.java
MrStahlfelge: Improved, that makes the methods more useful

MrStahlfelge:
> User creates a special AuthMessage which is like ErgoTransaction

This would introduce the problem that the AuthMessage is always the same. Hence anyone becoming aware of the signed message can authenticate to any place using the described scheme.

Apart from that, we could do it like described, if I understood everything correctly this adds only a reduction of ErgoTree under the current context (which makes sense). At the moment, this reduction is implemented in `AppKitProvingInterpreter` and can't be used for `Adress.getSigmaBoolean` though.

aslesarenko:
> AuthMessage is always the same

This problem can be solved independently by using DHTuple schema. I'm not sure, but @scalahub may suggest some ideas.

common/src/main/java/org/ergoplatform/appkit/ErgoAuthUtils.java
MrStahlfelge: Changed

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
MrStahlfelge: It is this step from the EIP:

> 6) When the user agrees, the wallet app **adds some own bytes** to the obtained message from ErgoAuthRequest,
> signs it  and sends the signed message to the ErgoAuthRequest's replyToUrl.

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
MrStahlfelge: Done

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
MrStahlfelge: Done

appkit/src/test/scala/org/ergoplatform/appkit/ErgoAuthSpec.scala
aslesarenko: I suggest to add this discription to the code comment (in test)

MrStahlfelge: done

aslesarenko (approved):

PR #164, opened 2022-04-19
Added Node Wallet support to BlockchainDataSource, along with a simple BoxLoader

K-Singh:
Would like a review, because implementing IUnspentBoxesLoader feels wrong due to needing an Address. Also no way to paginate boxes from the WalletAPI as far as I can tell, so had to pre-load all boxes which may be inefficient. I used to use `ctx.getWallet` quite a bit to get wallet boxes, so would like this functionality to still be supported some how.

### DIFF
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainDataSource.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BlockchainDataSource.java
@@ -72,6 +72,31 @@ public interface BlockchainDataSource {
     List<InputBox> getUnconfirmedUnspentBoxesFor(Address address, int offset, int limit);
 
     /**
+     * Get all unspent boxes in the current node wallet, without unconfirmed boxes
+     * @return List of input boxes representing all known
+     */
+    List<InputBox> getUnspentWalletBoxes();
+    /**
+     * Get all unspent boxes in the current node wallet, without unconfirmed boxes
+     * @param minConfirmations Minimum number of confirmations for a box to be included in the returned list
+     * @param minInclusionHeight Minimum creation height for a box to be included in the returned list
+     * @return List of input boxes representing all known
+     */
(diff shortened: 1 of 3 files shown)


MrStahlfelge:
Can you fix the tests?

MrStahlfelge:
We removed the wallet functionality from blockchain datasource completely: It is a functionality directly derived from restrictions of the node (it is not able to fetch unspend wallets for all addresses, but only for the own wallet), and it is not something the blockchain data is defining.

To work around this restriction without using the explorer you could introduce a NodeWalletDatasource implementing the `getUnspentBoxesFor` methods in a way that they actually return the node wallet's boxes.

However, I would completely dismiss this and work with the using an own loader only.

K-Singh:
> Can you fix the tests?

Yeah, didn't catch those before, will get a fix in tomorrow.

>We removed the wallet functionality from blockchain datasource completely: It is a functionality directly derived from restrictions of the node (it is not able to fetch unspend wallets for all addresses, but only for the own wallet), and it is not something the blockchain data is defining.

>To work around this restriction without using the explorer you could introduce a NodeWalletDatasource implementing the getUnspentBoxesFor methods in a way that they actually return the node wallet's boxes.

>However, I would completely dismiss this and work with the using an own loader only.

Sure, I can create a NodeWalletDataSource specifically for this. I don't entirely understand what you mean by the last line though.

MrStahlfelge:
> Sure, I can create a NodeWalletDataSource specifically for this. I don't entirely understand what you mean by the last line though.

The following: Instead of adapting the data source, you can just use the loader to fetch the data from the node. Instead of

     allBoxes = dataSource.getUnconfirmedUnspentWalletBoxes();

you do

    allBoxes = (dataSource as NodeAndExplorerDataSourceImpl).getWalletApi().getUnconfirmedUnspentWalletBoxes();

No need to change the datasource itself, just use an own loader. You can do this in your project without changing appkit.

PR #170, opened 2022-05-17
Fix Mainnet and Testnet Address being equal

MrStahlfelge:
This commit fixes that Testnet and Mainnet addresses were reported as equal by `Address.equals()`.

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/Address.java
+++ b/common/src/main/java/org/ergoplatform/appkit/Address.java
@@ -21,6 +21,8 @@ import special.sigma.GroupElement;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.common.base.Objects;
+
 public class Address {
     private final String _base58String;
     private final byte[] _addrBytes;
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/ergoplatform/appkit/Address.java
```diff
@@ -245,13 +247,14 @@ public String toString() {
 
     @Override
     public int hashCode() {
-        return _address.hashCode();
+        return Objects.hashCode(_address.hashCode(), _address.networkPrefix());
```
aslesarenko, line 250: Hm, this should also be fixed in Sigma, it seems.

aslesarenko (approved):

aslesarenko (changes requested):
Please add equality test case.

MrStahlfelge:
There's already some equality tests present, e.g. https://github.com/ergoplatform/ergo-appkit/blob/95a2bcd34d4f19e4703a96baf51cd218dbb207b0/appkit/src/test/scala/org/ergoplatform/appkit/AddressSpec.scala#L40

More are added in #171

aslesarenko (approved):

PR #181, opened 2022-07-12
Check tokens balance before TX signing

aslesarenko:
Token balance is checked by AppkitProvingInterpreter.reduceTransaction. When inputs/outputs balance is changed, then the difference is checked against explicitly requested token burning and possible token minting.

This check is independent from how UnsignedErgoLikeTransaction is created and it serves as sanity/security check against possible bugs in transaction building.


lib-api/src/main/java/org/ergoplatform/appkit/UnsignedTransaction.java
```diff
@@ -31,4 +27,10 @@ public interface UnsignedTransaction extends Transaction {
      * Returns the change address associated with this unsigned transaction
      */
     ErgoAddress getChangeAddress();
+
+    /**
+     * Optional list of tokens requested for burning in this transaction.
```
MrStahlfelge, line 32: Would be good to make it a bit more clear that this returns the tokens marked to burn, not the actual burned tokens. 

Maybe another method name would be better, as this one can easily misunderstood to return the actual tokens burned.

Also, `@Nonnull` annotation would be good to add here to make clear that the optional list is not returning null, but empty when there are no tokens requested to burn.

common/src/main/java/org/ergoplatform/appkit/JavaHelpers.scala
```diff
@@ -524,6 +535,40 @@ object JavaHelpers {
...
+
+  type TokenColl = Coll[(Coll[Byte], Long)]
+
+  def checkAllTokensPositive(tokens: TokenColl): TokenColl = {
+    val invalidTokens = tokens.filter(_._2 <= 0)
+    require(invalidTokens.isEmpty, s"All token values should be > 0: ")
+    tokens
+  }
+
+  def subtractTokens(
```
MrStahlfelge, line 547: Ah comment would help here, especially that this method can take multiple list elements of the same token and combines them by token id would be helpful here.

appkit/src/test/scala/org/ergoplatform/appkit/AppkitProvingInterpreterSpec.scala
```diff
@@ -0,0 +1,189 @@
...
+            case e: TokenBalanceException =>
+              val cond1 = exceptionLike[TokenBalanceException]("Transaction tries to burn tokens when no burning was requested")
+              cond1(e) && e.tokensDiff.exists(t => t == (Colls.fromArray(token2._1), -token2._2))
+            case _ => false
+          }
+        )
+      }
+
+      // Transaction tries to burn tokens when no burning was requested
+      {
```
MrStahlfelge, line 111: Good to have here would be a test that test with the following inputs and outputs:
Inputs:
Box1: Token1, Amount1
Box2: Token1. Amount2

Outputs:
Box1: Token1, Amount1

So that we have a test that checks if grouping a token is implemented correct.

MrStahlfelge (approved):
A good addition, only a few improvement suggestions

lib-api/src/main/java/org/ergoplatform/appkit/UnsignedTransaction.java
aslesarenko: Improved description and added annotation.
The name is consistent with boxesToSpend, inputsToSpend. Also `ToBurn` means in future, and the transaction not yet signed.
Please suggest a better name if you have some in mind.

common/src/main/java/org/ergoplatform/appkit/JavaHelpers.scala
aslesarenko: done

appkit/src/test/scala/org/ergoplatform/appkit/AppkitProvingInterpreterSpec.scala
aslesarenko: done

lib-api/src/main/java/org/ergoplatform/appkit/UnsignedTransaction.java
MrStahlfelge: I think i is clear enough with the new JavaDoc!

MrStahlfelge (approved):

PR #185, opened 2022-07-21
Generic builder of ErgoValue for Scala

### LINKED ISSUE #184: Implement generic builder of ErgoValue for Scala
Construction of ErgoValue instances with complex types of value is problematic, because ErgoValue is implemented with Java types, which makes it easier to be used from Java.

This issue is to implement generic builder

```scala
val x: Coll[(Coll[(Byte, Long)], Coll[Byte])] = ...

val ev = ErgoValueBuilder.buildFor(x)
```

aslesarenko:
Closes #184 
Implementation of generic builder to create ErgoValue instances with valid Java types from Scala.

### DIFF
--- a/common/src/main/java/org/ergoplatform/appkit/ErgoType.java
+++ b/common/src/main/java/org/ergoplatform/appkit/ErgoType.java
@@ -93,4 +93,8 @@ public class ErgoType<T> {
     static public <A> ErgoType<Option<A>> optionType(ErgoType<A> tItem) {
         return new ErgoType<>(RType.optionRType(tItem._rtype));
     }
+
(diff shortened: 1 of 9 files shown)


MrStahlfelge (approved):
LGTM, but let's wait for feedback by CheeseEnthusiast if this covers all his use cases.

Only small improvement suggestion is adding a hint to ErgoValue and ErgoValueBuilder that the other builder class for the other language exists.

K-Singh:
Yeah this just about fixes most of the problems I faced. I found ways to do it manually but this is much much simpler and makes using ErgoValue's from Scala a lot easier.

PR #192, opened 2022-08-31
Migrate to sigma v5.0

aslesarenko:
This PR migrates the appkit code to use sigmastate v5.0


MrStahlfelge:
Can you check the compile errors?

MrStahlfelge (approved):
This looks good to me.

Does it make sense to add the change in `ColdErgoClient` constructor to MIGRATION file?

PR #193, opened 2022-09-01
BoxOperations: property to restrict the amount of input boxes

MrStahlfelge:
Recently several miners with small payout boxes expressed problems in the support channels sending ERG caused by too many input boxes. 

This expresses in the typical error indications:
* Error when submitting the transaction "too high cost"
* Very long loading time
* Timeout exceptions connecting Explorer and/or node

Too avoid all of these errors caused by too many needed input boxes, and to give a helpful error message to the users with a hint how many ERG and/or tokens can be sent with a certain set input boxes restriction, this adds an optional input box selection restriction to BoxOperations. When the amount of selected (not loaded) input boxes is exceeded, an exception is thrown that can be used to give the user helpful hints.

If the restriction is not set, the behaviour does not change.

### DIFF
--- a/lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
+++ b/lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
@@ -28,6 +28,7 @@ public class BoxOperations {
     private long feeAmount = MinFee;
     private IUnspentBoxesLoader inputBoxesLoader = new ExplorerApiUnspentLoader();
     private BoxAttachment attachment;
+    private int maxInputBoxesToSelect = 0;
 
     private static final long CHANGE_BOX_NANOERG = MinFee;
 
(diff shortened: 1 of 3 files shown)


lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -115,6 +116,21 @@ public BoxOperations withAttachment(@Nullable BoxAttachment attachment) {
...
 
+    public int getMaxInputBoxesToSelect() {
+        return maxInputBoxesToSelect;
+    }
+
+    /**
+     * @param maxInputBoxesToSelect if set greater than 0, {@link #loadTop()} will only select
+     *                              up to this number of input boxes to satisfy ERG and token amount
+     *                              needed and throws an {@link org.ergoplatform.appkit.InputBoxesSelectionException.InputBoxLimitExceededException}
+     *                              otherwise
```
aslesarenko, line 127: Please describe what the default value of 0 means.

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -406,6 +437,17 @@ public static CoveringBoxes getCoveringBoxesFor(long amountToSpend,
...
+                    else if (maxBoxesToSelect > 0 && selectedCoveringBoxes.size() >= maxBoxesToSelect) {
+                        List<ErgoToken> remainingTokenList = tokensRemaining.getRemainingTokenList();
+                        throw new InputBoxesSelectionException.InputBoxLimitExceededException(
+                            "Input box limit exceeded, could not cover " + remainingAmountToCover +
+                                " nanoERG and " + remainingTokenList.size() + " tokens.",
+                            remainingAmountToCover,
+                            remainingTokenList,
+                            maxBoxesToSelect
+                        );
+                    }
```
aslesarenko, line 450: What if maxBoxesToSelect == 0?

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
MrStahlfelge: Done

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
MrStahlfelge: Added a comment

aslesarenko (approved):

PR #197, opened 2022-10-24
New TransactionsApi endpoints added in ergoplatform/ergo#1869

MrStahlfelge:
No tests added as these endpoints are not supported by most nodes out there (yet)

### DIFF
--- a/java-client-generated/src/main/java/org/ergoplatform/restapi/client/TransactionsApi.java
+++ b/java-client-generated/src/main/java/org/ergoplatform/restapi/client/TransactionsApi.java
@@ -78,6 +78,37 @@ public interface TransactionsApi {
   );
 
   /**
+   * Get unconfirmed transaction from pool. Available from node version 4.0.105 and up. Older
+   * versions will return a 404 error, as will newer versions when the transaction is not in
+   * mempool.
+   *
+   * @param txId ID of a transaction in question
+   * @return Ergo Transaction
+   */
+  @GET("transactions/unconfirmed/byTransactionId/{txId}")
+  Call<ErgoTransaction> getUnconfirmedTransactionById(
+        @retrofit2.http.Path("txId") String txId
+  );
+
+  /**
+   * Finds unconfirmed transactions by ErgoTree hex of one of its output or input boxes
+   * (if present in UtxoState). Available from node version 4.0.105 and up. Older
+   * versions will return a 404 error.
+   *
+   * @param ergoTreeHex ErgoTree hex representation with surrounding quotes ("0008cd...")
+   * @param limit The number of items in list to return (optional, default to 50)
+   * @param offset The number of items in list to skip (optional, default to 0)
+   * @return Call&lt;Transactions&gt;
+   */
+  @POST("transactions/unconfirmed/byErgoTree")
+  @Headers({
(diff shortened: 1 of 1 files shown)


java-client-generated/src/main/java/org/ergoplatform/restapi/client/TransactionsApi.java
```diff
@@ -77,6 +77,37 @@ Call<Transactions> getUnconfirmedTransactions(
         @retrofit2.http.Query("limit") Integer limit                ,     @retrofit2.http.Query("offset") Integer offset                
   );
 
+  /**
+   * Get unconfirmed transaction from pool. Available from node version 4.0.105 and up. Older
+   * versions will return a 404 error, as will newer versions when the transaction is not in
+   * mempool.
```
aslesarenko, line 83: Using 404 error doesn't looks like a good idea to me. 
It is better to return successfully even if there was no transaction in the pool.
Otherwise hard to distinguish between network problem and just absence of tx.
The first should be troubleshooted, the later maybe wait or something.

aslesarenko (changes requested):

java-client-generated/src/main/java/org/ergoplatform/restapi/client/TransactionsApi.java
pragmaxim: We just agreed with @kushti that I change the Node's `RejectionHandler` to return `400` instead of `404` for querying non-existing endpoint ... as currently we use `404` when we ask for a resource like `TX` and it is not present... We cannot return `200` as there is nothing to put into Body. And there is a lot of these `Option[T]` endpoints

MrStahlfelge: That is good news. However, the problem will persist that old nodes still return 404 as do nodes when the transaction does not exist. This problem will become less important in the future when newer node versions are spreading.

aslesarenko: > agreed with @kushti that I change the Node's `RejectionHandler` to return `400`

@pragmaxim so this will fix all the enpoints, not just this specific one?

pragmaxim: @aslesarenko yes it is done at RejectionHandler level https://github.com/ergoplatform/ergo/pull/1882/files#diff-7f87a6eb6aa8d2abbbb020080fa251cfb6b36afa976ab2c869766ffe92fe42fdR34

aslesarenko: @MrStahlfelge so, with that difference on 400 and 404, are any changes need to be done here. Maybe update doc comments?

MrStahlfelge: No changes to the doc needed: New node versions support the endpoint and will return 404 if transaction is not available or 2xx if it is. Old endpoints will return 404 in any case. We only solved 404 clashing the problem for future endpoints to be added.

aslesarenko (approved):

PR #198, opened 2022-10-24
Add methods to access spent boxes and save to registers

MrStahlfelge:
Mgpai discovered that retrieving spent boxes to use them in box registers is currently not straightforward with appkit. This PR adds methods to make this task more easy for client code:

* `BlockchainDataSource#getBoxByIdWithSpent`
* `InputBox#toErgoValue`
* `InputBoxImpl` constructor for Explorer OutputInfo (used by `BlockchainDataSource#getBoxByIdWithSpent`)

### DIFF
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
@@ -157,6 +157,12 @@ public class NodeAndExplorerDataSourceImpl implements BlockchainDataSource {
     }
 
     @Override
+    public InputBox getBoxByIdWithSpent(String boxId) {
+        OutputInfo outputInfo = executeCall(explorerApi.getApiV1BoxesP1(boxId));
+        return new InputBoxImpl(outputInfo);
+    }
+
+    @Override
     public String sendTransaction(SignedTransaction tx) {
         ErgoLikeTransaction ergoTx = ((SignedTransactionImpl) tx).getTx();
         List<ErgoTransactionDataInput> dataInputsData =
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
@@ -145,19 +145,28 @@ public class NodeAndExplorerDataSourceImpl implements BlockchainDataSource {
     }
 
     @Override
-    public InputBox getBoxById(String boxId) {
-        ErgoTransactionOutput boxData = executeCall(nodeUtxoApi.getBoxById(boxId));
+    public InputBox getBoxById(String boxId, boolean findInPool, boolean findInSpent) {
+        if (findInSpent && !findInPool) {
+            return getBoxByIdExplorer(boxId);
+        } else if (!findInSpent) {
+            return getUnspentBoxByIdNode(boxId, findInPool);
+        } else {
+            // find in spent and find in pool => try node first and fall back to explorer
+            try {
+                return getUnspentBoxByIdNode(boxId, findInPool);
+            } catch (Throwable t) {
+                return getBoxByIdExplorer(boxId);
+            }
+        }
+    }
+
+    private InputBox getUnspentBoxByIdNode(String boxId, boolean findInPool) {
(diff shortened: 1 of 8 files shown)


lib-impl/src/main/java/org/ergoplatform/appkit/impl/ScalaBridge.scala
```diff
@@ -170,33 +171,34 @@ object ScalaBridge {
...
+          .value(box.value)
+          .ergoTree(ErgoAlgos.encode(TreeSerializer.serializeErgoTree(box.ergoTree)))
+          .assets(assets.map(asset => new AssetInstanceInfo().tokenId(asset.getTokenId).amount(asset.getAmount)))
+          .additionalRegisters(regs)
+          .creationHeight(box.creationHeight)
+          .transactionId(box.transactionId)
+          .index(box.index)
+      out
+    }
+  }
```
aslesarenko, line 201: Does the `to` and `from` mutually reverse. They are maybe not because of that `sortBy`.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
```diff
@@ -156,6 +156,12 @@ public InputBox getBoxByIdWithMemPool(String boxId) {
         return new InputBoxImpl(boxData);
     }
 
+    @Override
+    public InputBox getBoxByIdWithSpent(String boxId) {
+        OutputInfo outputInfo = executeCall(explorerApi.getApiV1BoxesP1(boxId));
+        return new InputBoxImpl(outputInfo);
+    }
```
aslesarenko, line 163: What is the difference with getBoxById other than different types involved?

MrStahlfelge: `getBoxById` only returns unspent boxes. It would be better to rename it, but that would break a lot of client code.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ScalaBridge.scala
MrStahlfelge: Good catch, the index needs to be set

lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
aslesarenko: Then maybe give it a more descriptive name or add parameter
`InputBox getBoxById(String boxId, boolean searchInSpentBoxes)`

lib-impl/src/main/java/org/ergoplatform/appkit/impl/ScalaBridge.scala
MrStahlfelge: Done

lib-impl/src/main/java/org/ergoplatform/appkit/impl/NodeAndExplorerDataSourceImpl.java
MrStahlfelge: What do you suggest? To be backwards compatible, we need these two:

`getBoxById(String boxId)` -> only unspent boxes
`getBoxById(String boxId, boolean searchInSpentBoxes)` -> either only unspent boxes or search in spent boxes as well

this breaks down into the two options that are available now as well, only that instead of `getBoxByIdWithSpent` you need to call `getBoxById(String boxId, true)`

aslesarenko: I think the meaning of getBoxById(String boxId, boolean searchInSpentBoxes) is more clear than getBoxByIdWithSpent (that "WithSpent" is not clear). At the same time I don't like overloading methods.
An alternative name `findBoxByIdIncludingSpent`, which suggest that there will be some `find` which includes the spent boxes.

aslesarenko: Мaybe just one generic method `findBoxById(id, findInPool = true, findInSpent = true)`
or even introduce a class `SearchParameters(findInUtxo: Boolean, findInPool:Boolean, findInSpent: Boolean)` 
With this you can find everywhere except Utxo, which is not possible now.

MrStahlfelge: Done with the parameter flags.

PR #200, opened 2022-10-30
Add methods to construct Coll from Java and use in ErgoValue more easy

MrStahlfelge:
This adds method to use `Coll` more easily from Java

### DIFF
--- a/appkit/src/test/scala/org/ergoplatform/appkit/ErgoValueTest.java
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/ErgoValueTest.java
@@ -20,9 +20,17 @@ public class ErgoValueTest {
         ErgoValue<Short> shortErgoValue = ErgoValue.of((short) 1);
         ErgoValue<BigInt> bigIntErgoValue = ErgoValue.of(BigInteger.ZERO);
         ErgoValue<Coll<Byte>> byteArrayErgoValue = ErgoValue.of(new byte[] {0, 1, 2});
+        ErgoValue<Coll<Short>> shortArrayErgoValue = ErgoValue.of(new short[] {0, 1, 2});
+        ErgoValue<Coll<Integer>> intArrayErgoValue = ErgoValue.of(new int[] {0, 1, 2});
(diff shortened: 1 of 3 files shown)


appkit/src/test/scala/org/ergoplatform/appkit/ErgoValueTest.java
```diff
@@ -34,7 +42,11 @@ public void testTypeDeclarations() {
         assertEquals(true, booleanValue);
         assertEquals(1, byteValue);
         assertEquals(1, shortValue);
+        assertFalse(booleanFromCollValue);
         assertEquals(0, byteFromCollValue);
+        assertEquals(0, shortFromCollValue);
+        assertEquals(0, intFromCollValue);
+        assertEquals(0, longFromCollValue);
```
aslesarenko, line 49: I think `0` is not good test vector value, as it often may mean `uninitialized` value.

common/src/main/java/org/ergoplatform/appkit/JavaHelpers.scala
```diff
@@ -475,6 +475,14 @@ object JavaHelpers {
...
   }
 
+  def collFrom(arr: Array[Long]): Coll[Long] = Colls.fromArray(arr)
+
+  def collFrom(arr: Array[Int]): Coll[Int] = Colls.fromArray(arr)
+
+  def collFrom(arr: Array[Boolean]): Coll[Boolean] = Colls.fromArray(arr)
+
+  def collFrom(arr: Array[Short]): Coll[Short] = Colls.fromArray(arr)
+
```
aslesarenko, line 485: Of course it is better to have Iso[Array[T], Coll[T]], but this is also good. :)

aslesarenko (approved):
Looks good, but please see comments.

common/src/main/java/org/ergoplatform/appkit/JavaHelpers.scala
MrStahlfelge: The iso is not vital as from Java values of Coll can be accessed with coll.apply(i) - but it is not possible to construct a `Coll`

PR #73, opened 2021-04-06
Add backwards compatible 'formatJson' option to the toJson methods

wclaeys:
Fixes #27

PR #213, opened 2022-12-05
Add json output for unsigned transactions

MrStahlfelge:


lib-api/src/main/java/org/ergoplatform/appkit/UnsignedTransaction.java
```diff
@@ -37,4 +37,16 @@ public interface UnsignedTransaction extends Transaction {
...
     List<ErgoToken> getTokensToBurn();
+
+    /**
+     * Json representation of this transaction.
+     */
+    @Nonnull
+    String toJson(boolean prettyPrint);
+
+    /**
+     * Json representation of this transaction.
```
aslesarenko, line 48: please add description of parameters

aslesarenko (changes requested):

lib-api/src/main/java/org/ergoplatform/appkit/UnsignedTransaction.java
MrStahlfelge: Are you referring to prettyPrint?

aslesarenko: yes, all parameters, the meaning is not intuitive

MrStahlfelge: Tbh I don't understand them, I just followed the implementation for `SignedTransaction`. The parameters were introduced in #73. Maybe @wclaeys can shed some light.

aslesarenko: hi, surprised you found it problematic :)

aslesarenko (approved):

PR #214, opened 2023-02-19
Eip4 update

mgpai22:
Update to allow collection token minting in [updated eip4](https://github.com/anon-real/eips/blob/master/eip-0004.md)

Must be merged with master 5.0 since develop currently has breaking changes

### DIFF
--- a/lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
+++ b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
@@ -39,11 +39,14 @@ class OutBoxBuilderImpl(_txB: UnsignedTransactionBuilderImpl) extends OutBoxBuil
     val tokenNumOfDecVal = token.getMintingBoxR6
     _registers ++= Array(tokenNameVal, tokenDescVal, tokenNumOfDecVal)
 
-    // optional registers, but either all of them or none
-    if (token.getMintingBoxR7 != null && token.getMintingBoxR8 != null) {
-      _registers ++= Array(token.getMintingBoxR7, token.getMintingBoxR8)
-      if (token.getMintingBoxR9 != null)
-        _registers += token.getMintingBoxR9
+    if(token.getMintingBoxR7 != null){
+      _registers ++= Array(token.getMintingBoxR7)
+    }
+    if(token.getMintingBoxR8 != null){
+      _registers ++= Array(token.getMintingBoxR8)
+    }
+    if(token.getMintingBoxR9 != null){
+      _registers ++= Array(token.getMintingBoxR9)
     }
 
     _tokens += token
--- a/appkit/src/test/scala/org/ergoplatform/appkit/examples/DataInputsSpec.scala
+++ b/appkit/src/test/scala/org/ergoplatform/appkit/examples/DataInputsSpec.scala
@@ -34,6 +34,9 @@ class DataInputsSpec extends PropSpec with Matchers
         ConstantsBuilder.create().item("expectedDataInputPropBytes", dataInputPropBytes).build(), inputBoxScript
       )).build().convertToInputWith(dummyTxId, 0)
 
+      input.getTransactionId shouldBe dummyTxId
+      input.getTransactionIndex shouldBe 0
+
       val txB = ctx.newTxBuilder()
 
       val dummyOutput = txB.outBoxBuilder()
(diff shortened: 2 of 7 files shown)


lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
```diff
@@ -39,11 +39,14 @@ class OutBoxBuilderImpl(_txB: UnsignedTransactionBuilderImpl) extends OutBoxBuil
...
-      if (token.getMintingBoxR9 != null)
-        _registers += token.getMintingBoxR9
+    if(token.getMintingBoxR7 != null){
+      _registers ++= Array(token.getMintingBoxR7)
+    }
+    if(token.getMintingBoxR8 != null){
+      _registers ++= Array(token.getMintingBoxR8)
+    }
+    if(token.getMintingBoxR9 != null){
+      _registers ++= Array(token.getMintingBoxR9)
```
aslesarenko, line 49: There is a potential problem here. When R8 == null, then the value of token.getMintingBoxR9 will be added after R7 and will lend into R8. There will be a lot of confusion and hard to debug.
So, R9 cannot be added is R8 is null. 
The same for R7 (when it is null, R8 and R9 cannot be added).
This constrants better be checked early on.

aslesarenko (changes requested):

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
mgpai22: Good point, please see the fix.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
```diff
@@ -41,14 +41,18 @@ class OutBoxBuilderImpl(_txB: UnsignedTransactionBuilderImpl) extends OutBoxBuil
...
+        _registers ++= Array(token.getMintingBoxR8)
+
+
+        if(token.getMintingBoxR9 != null){
+          _registers ++= Array(token.getMintingBoxR9)
+        }
+
+      }
+
+    }
```
aslesarenko, line 55: @mgpai22, please think about full validation. Whati token.getMintingBoxR7 == null, but token.getMintingBoxR8 or token.getMintingBoxR9 are not. This shouldn't be silently swallowed, it is better to throw a IllegalArgumentException.

aslesarenko (changes requested):

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
mgpai22: Good point, the old code caused bugs for me before while testing the modified eip4 standard since there weren't exceptions thrown.

aslesarenko (changes requested):
Please add test cases in Eip4TokenBuilderTest for newly added ARTWORK_COLLECTION.
Both positive and negative to cover error branches.

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
```diff
@@ -39,13 +39,33 @@ class OutBoxBuilderImpl(_txB: UnsignedTransactionBuilderImpl) extends OutBoxBuil
...
+        }
+      }
+
+    } else {
+      if(token.getMintingBoxR8 != null){
+        throw new IllegalArgumentException("R8 cannot exist without R7")
+      }
+      if(token.getMintingBoxR9 != null){
+        throw new IllegalArgumentException("R9 cannot exist without R7")
+      }
```
aslesarenko, line 65: That is a lot of nesting. It would be easier to add nulls to _registers and then just check finally that _registers don't have nulls. And throw otherwise. This will be easier to cover with tests.

aslesarenko (changes requested):

aslesarenko (changes requested):
Please fix the errors

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
mgpai22: This has some logic error which fails tests, please see how I changed implementation to remove nested statements.

PR #224, opened 2023-05-20
Transactions API: Fix getUnconfirmedTransactionsByErgoTree text/plain mode

MrStahlfelge:
In past versions, it was needed to wrap the ergotree param in quotes when ScalarsConverterFactory was used for Retrofit. This is fixed and the workaround needs to be removed when used.


MrStahlfelge:
Closed: node does not accept text/plain body

PR #225, opened 2023-05-27
Upgrade to sigma v5.0.8

aslesarenko:
In this PR:
- [x] the code is upgraded to sigma v5.0.8
- [x] AppkitProvingInterpreter and related code moved to Sigma sdk module
- [x] `$.MODULE$` removed where possible
- [x] imports optimized

No tests changed, despite many changes the backward compatibility is preserved.


MrStahlfelge (commented):
This looks good overall, but the package change for ErgoToken and ErgoId will make changes needed in client code as these are widely used in public methods. Should be stated in the migration file or by adding deprecated empty subclasses (`org.ergoplatform.appkit.ErgoToken extends org.ergoplatform.sdk.ErgoToken`)

aslesarenko:
@MrStahlfelge, tried to add empty classes. This hasn't ended up well, because class duplication somehow confuses Scala 2.11, but not Scala 2.12, 2.13. Will also add migration note in release notes.

MrStahlfelge (approved):

PR #239, opened 2024-09-06
Removed 2 libraries with vulnerabilities

satsen:
The versions of Guava and Commons IO used in Appkit have vulnerabilities. Even if the vulnerable code might not be invoked (I didn't check), it's a bad look because all users of the library receive warnings. These libraries are not used much at all so removing them is the best solution.


kushti (approved):

common/src/test/scala/org/ergoplatform/appkit/SecretStorageSpec.scala
```diff
@@ -69,7 +70,7 @@ class SecretStorageSpec extends AnyPropSpec with Matchers with ScalaCheckDrivenP
   }
 
   def withTempDir(block: File => Unit): Unit = {
-    val dir = Files.createTempDir()
+    val dir = Files.createTempDirectory("").toFile
```
aslesarenko, line 73: Is this method available in Java 7?

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
```diff
@@ -63,8 +61,9 @@ public static BoxOperations createForSenders(List<Address> senders, BlockchainCo
      */
     public static BoxOperations createForEip3Prover(ErgoProver senderProver, BlockchainContext ctx) {
         List<Address> eip3Addresses = senderProver.getEip3Addresses();
-        checkState(eip3Addresses.size() > 0,
-            "EIP-3 addresses are not derived in the prover (use ErgoProverBuilder.withEip3Secret)");
+        if (eip3Addresses.isEmpty()) {
+            throw new IllegalStateException("EIP-3 addresses are not derived in the prover (use ErgoProverBuilder.withEip3Secret)");
+        }
```
aslesarenko: Why not add checkState to the InternalUtil?

lib-impl/src/main/java/org/ergoplatform/appkit/impl/OutBoxBuilderImpl.scala
```diff
@@ -78,7 +76,9 @@ class OutBoxBuilderImpl(_txB: UnsignedTransactionBuilderImpl) extends OutBoxBuil
   }
 
   override def build: OutBox = {
-    checkState(_contract != null, "Contract is not defined": Any)
+    if (_contract == null) {
+      throw new IllegalStateException("Contract is not defined")
+    }
```
aslesarenko: Suggest to add method to InternalUtil.

common/src/main/java/org/ergoplatform/appkit/Address.java
```diff
@@ -86,7 +85,7 @@ public NetworkType getNetworkType() {
      * @throws IllegalArgumentException if this instance is not P2PK address
      */
     public P2PKAddress asP2PK() {
-        checkArgument(isP2PK(), "This instance %s is not P2PKAddress", this);
+        InternalUtil.checkArgument(isP2PK(), String.format("This instance %s is not P2PKAddress", this));
```
aslesarenko: The previous code avoided String.format to be called when no errors.

common/src/main/java/org/ergoplatform/appkit/Address.java
```diff
@@ -100,7 +99,7 @@ public P2PKAddress asP2PK() {
      * @throws IllegalArgumentException if this instance is not P2S address
      */
     public Pay2SAddress asP2S() {
-        checkArgument(isP2S(), "This instance %s is not Pay2SAddress", this);
+        InternalUtil.checkArgument(isP2S(), String.format("This instance %s is not Pay2SAddress", this));
```
aslesarenko: Same as above.

common/src/main/java/org/ergoplatform/appkit/Address.java
```diff
@@ -279,7 +278,7 @@ public String toString() {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(_address.hashCode(), _address.networkPrefix());
+        return Objects.hash(_address.hashCode(), _address.networkPrefix());
```
aslesarenko, line 281: What is the motivation for this change?

aslesarenko (changes requested):

common/src/test/scala/org/ergoplatform/appkit/SecretStorageSpec.scala
satsen: Yes

lib-api/src/main/java/org/ergoplatform/appkit/BoxOperations.java
satsen: Felt unnecessary for 2 uses, and in my opinion this code is clearer. But I can add it if you want. @aslesarenko

Edit: Added

common/src/main/java/org/ergoplatform/appkit/Address.java
satsen: Is it it ok to just use a string concatenation then

common/src/main/java/org/ergoplatform/appkit/Address.java
satsen: com.google.common.base.Objects.hashCode is a method in the Guava library.
java.util.Objects.hash is a method in the JDK.

Their implementation is the same. The motivation is that I removed Guava.

aslesarenko (approved):

PR #241, opened 2024-10-10
Removed CLI feature

satsen:
This feature is only used by https://github.com/ergoplatform/ergo-tool, there is even a help message that mentions "ergotool", so it is appropriate to move it there instead. It does not make sense to include a CLI framework in appkit.

After removing all code related to this feature, there was only one Scala class left in the appkit module, so I rewrote it in Java to make it tidier.


kushti (approved):
@ergomorphic please take a look as well

kushti (approved):

